PR #22, opened 2016-07-11
speedup SDF and openCL "double call' wrapper

tpietzsch:


skalarproduktraum:
Thanks, the speedup is amazing!

tpietzsch:
Yay!

PR #24, opened 2016-07-13
SceneryDefaultJPanelApplication and JPanel example

kephale:


skalarproduktraum:
Oooh, this is great stuff, thanks @kephale 👍

PR #26, opened 2016-07-14
WIP: Bounding boxes

skalarproduktraum:
* Nodes now have a boundingBoxCoords property that stores their bounding box. This
  property gets set e.g. by HasGeometry upon model loading
* there is a new BoundingBox object which makes use of the aforementioned property. This
  object will display a customizable coordinate grid with axis labels

### DIFF
--- a/src/main/kotlin/scenery/BoundingBox.kt
+++ b/src/main/kotlin/scenery/BoundingBox.kt
@@ -1,6 +1,8 @@
 package scenery
 
 import cleargl.GLVector
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 import scenery.rendermodules.opengl.OpenGLShaderPreference
 import java.util.*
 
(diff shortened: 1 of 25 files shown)


skalarproduktraum:
Progress:
screen shot 2016-07-15 at 20 24 33

Screen-space anti-aliased grid lines, configurable grid color, labels, number of lines... status as of bb53dc3

royerloic:
Ulrik is on fire!!

skalarproduktraum:
Finalized so far. A BoundingBox will evaluate the node it's attached to for it's node.boundingBoxCoords. It'll however _not_ recurse into the node's children. Use via

```
val bb = BoundingBox()
bb.node = TheNodeIWantToSeeTheBoundingBoxOf()
``

The BoundingBox will attach itself then to the node as a child, as to correctly capture it's transformations.

PR #45, opened 2017-01-06
Use the 3.1.1 release of LWJGL

ctrueden:
The `3.1.1-SNAPSHOT` builds are [gone now](http://maven.imagej.net/service/local/artifact/maven/redirect?r=sonatype-snapshots&g=org.lwjgl&a=lwjgl&v=3.1.1-SNAPSHOT&e=pom), due to the snapshot retention policy of the OSS Sonatype Snapshots repository. Using `3.1.1` instead is stable and fixes the build.


skalarproduktraum:
This is incorporated in 98867ac3466d92e3fea0079b5f0569c878f37a3e -- thanks for the PR! 👍

PR #70, opened 2017-06-15
Normal compression and alchemy ssao

aryaman1994:
This version contains implementation of Alchemmy SSAO algorithm, and compression of Normal Vectors using Octahedron and Spherical unit vectors, for the Vulkan Renderer. The usage of z component in the SSAO algorithm as Bias Scale needs to be verified and corrected.

### DIFF
--- a/src/main/resources/graphics/scenery/backends/vulkan/shaders/DeferredLighting.frag
+++ b/src/main/resources/graphics/scenery/backends/vulkan/shaders/DeferredLighting.frag
@@ -37,13 +37,19 @@ layout(set = 2, binding = 0, std140) uniform LightParameters {
 
 layout(set = 3, binding = 0, std140) uniform ShaderParameters {
 	int debugBuffers;
-	int activateSSAO;
+	int SSAO_Options;
 	int reflectanceModel;
 	float ssaoDistanceThreshold;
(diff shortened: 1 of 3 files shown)


src/main/resources/graphics/scenery/backends/vulkan/shaders/DeferredLighting.frag
```diff
@@ -231,27 +305,27 @@ void main()
...
-		}
-		// normal
-		if(textureCoord.x > 0.5 && textureCoord.y > 0.5) {
-			FragColor = vec4(Normal, 1.0f);
-		}
-		// position
-		if(textureCoord.x < 0.5 && textureCoord.y > 0.5) {
-			FragColor = vec4(FragPos, 1.0f);
-		}
+        if(textureCoord.x < 0.25 && textureCoord.y < 0.5 ) {
```
skalarproduktraum: It looks like the indentation here is not as it should be, could you check that please? Thanks!

skalarproduktraum:
Thanks @aryaman1994, awesome work! 👍

aryaman1994:
Thanks @skalarproduktraum. Hope for many more 👍

PR #75, opened 2017-06-22
Fix examples

skalarproduktraum:
* organises examples into basic examples, advanced examples, stress tests, and experimental cluster examples
* fixes various issues with existing examples


kephale:
ORC6.obj is not provided with the demo files
Default lighting on Sponza is really dark (with OpenGL on my OS X machine)
powerplant.obj is not provided with the demo files
rungholt.obj is not provided with the demo files

skalarproduktraum:
New models available at https://ulrik.is/scenery-demo-models.zip

PR #85, opened 2017-07-11
Updating to recent scijava-common that solves downstream ops bug

kephale:


pom.xml
```diff
@@ -293,7 +294,6 @@
         <dependency>
             <groupId>org.scijava</groupId>
             <artifactId>scijava-common</artifactId>
-            <version>2.61.0</version>
```
skalarproduktraum: This should need a `<version>${scijava-common.version}</version>` here, not?

kephale:
Worked without it for me actually. I ran the examples with the modified version, but happy to add.

pom.xml
imagejan: No need to specify versions that are managed in the parent POM. You can leave away the `<version>` tag entirely.

skalarproduktraum:
Maybe it wouldn't hurt to make it explicit :-)

pom.xml
skalarproduktraum: Thanks @imagejan, I'll merge it as-is then 👍

PR #88, opened 2017-07-18
Make junit test scope with default version

ctrueden:


skalarproduktraum:
Thanks for the PR, @ctrueden. Though some time ago, I added JUnit to the normal scope, as all the examples are JUnit tests, and I remember it caused trouble when I moved JUnit to the test scope. Any advise?

skalarproduktraum:
Okay, I just tested this locally, both command line and IntelliJ, and it didn't seem to cause any ill effects. I'll merge this (and the PR at scenerygraphics/spirvcrossj) as soon as Travis is done. Thanks again, @ctrueden! 👍

ctrueden:
> I just tested this locally, both command line and IntelliJ, and it didn't seem to cause any ill effects.

Yeah, it worked for me from the CLI as well. That's why I filed it. Glad it works for you, too.

PR #92, opened 2017-07-31
Fix mesh upside down after window resize - issue fix #90

### LINKED ISSUE #90: Mesh gets upside down when the window is resized (using JavaFX)
I'm running this [example](https://github.com/ssinhaleite/MarchingCubes/blob/master/src/tests/JavaFXApplication.java) and when the window is resized, the mesh turns upside down.

ssinhaleite:
I noticed that when the panel is created, the Y axis is inverted, but, when the window is resized, the panel is updated, but the Y axis is not inverted. Maybe this is not the right solution, but this fixed the problem.


skalarproduktraum:
Hi @ssinhaleite,

thanks for the PR! I fixed this now at a deeper level, in SceneryPanel directly, see ae3dc6bd621a8dba05e62c554364781e28ec66dd

Have a good weekend!

PR #100, opened 2017-10-12
Update the mesh bounding box generation

ssinhaleite:


skalarproduktraum:
Hi @ssinhaleite,

thanks for the PR! Both problems are however already fixed in #101, which I'll merge later today or tomorrow. Also, using `FloatBuffer.array()` does not always work, as in the case of NIO buffers, they are not backed by an array and that function will throw an exception.

I'll close this for the moment, but please reopen in case I have missed something!

cheers and thanks :)

PR #114, opened 2017-11-16
Add AMD Radeon R9 M370X to supported GPUs

ctrueden:
As requested by @skalarproduktraum [here](https://gitter.im/scenerygraphics/SciView?at=5a0d9baf0cef4fed2ebd5860).


skalarproduktraum:
@ctrueden unfortunately, I made a mistake... so there is actually no M930X, it is a typo and should actually say M370X, which is what you just added. Could you just delete the line after your addition, then I'd merge. Thanks and sorry for my uncoordinated fingers!

ctrueden:
@skalarproduktraum I fixed the commit! I also split out the EOL newline change into its own commit. (Previously, GitHub's online editor added the EOL newline automatically, which is why it was mixed in.)

I was wondering though about the "Strato Pro" in parentheses. I do not know what that means, so I didn't remove it, but I see no mention of "Strato Pro" in the system configuration of my machine. So I was wondering what that notation indicates, whether it is accurate, and how to check on a machine what should be noted there?

skalarproduktraum:
@ctrueden Thanks!

About the Strato Pro: That's the code name of the chip or the chips architecture. Wikipedia has an extensive list: https://en.wikipedia.org/wiki/List_of_AMD_graphics_processing_units and https://en.wikipedia.org/wiki/List_of_Nvidia_graphics_processing_units

I'd like to add them there as especially the sneaky b... erm, great marketers, from Nvidia like to up the series number, but still use an older gen chip, especially for mobile GPUs. The GT630 is a good example: 600 series, which would usually mean Kepler architecture, but in that case with an older Fermi core.

PR #121, opened 2017-12-06
Point cloud fix

kephale:


pom.xml
```diff
@@ -307,7 +307,7 @@
             <artifactId>ui-behaviour</artifactId>
             <version>1.2.0</version>
         </dependency>
-
+
```
skalarproduktraum: Sorry for being picky, but could you remove that?

skalarproduktraum:
Thank you, @kephale!

PR #130, opened 2017-12-20
Do not set prefWidth/Height in renderer

hanslovsky:
The renderer should not modify properties of the scenery panel, thus remove all calls to `panel.{pref,min}{Width,Height}` in `OpenGLRenderer.kt`. Still, the minimum width and height need to be larger than zero to avoid zero length rendering buffers. This is handled in the `SceneryPanel` constructor.


skalarproduktraum:
Thank you, @hanslovsky!

PR #144, opened 2018-04-22
first sketch

elect86:
All short-lived allocations are using a memory pool that is being allocated once at the start and reset/cleared at every refresh (it shall be done also at every resize, because if the user resizes a lot without leaving the button mouse you'll run out of it)

Added some higher vk level calls, the return `VkResult` int will be checked automatically in debug mode under the hood. 

Set `VULKAN_NO_EXCEPTIONS` accordingly if you want at errors the exception being thrown or not 

Switched many `if-else` to `when` for better code readability

Using operators `set`, `get`, `invoke`

Using typealias for vk resources

And many small other fixes


skalarproduktraum:
Oh, this is quite exciting, I like how the code becomes quite a bit cleaner. Do the examples still work and is the Kotlin EAP version strictly necessary?

elect86:
Thanks, I appreciate your early feedbacks :)

I relied the whole time on the `TexturedCubeExample` (with resizing)

No, the EAP is not necessary, I'll update sooner the whole project to kt .40

pom.xml
```diff
@@ -503,6 +509,17 @@
             <artifactId>ffmpeg-platform</artifactId>
             <version>3.4.1-1.4</version>
         </dependency>
+        <dependency>
```
skalarproduktraum: These guys are already there

pom.xml
```diff
@@ -519,6 +536,21 @@
...
+            <releases>
+                <enabled>true</enabled>
+            </releases>
+            <snapshots>
+                <enabled>false</enabled>
+            </snapshots>
+            <id>bintray.kotlin.eap</id>
+            <name>Bintray Kotlin EAP Repository</name>
+            <url>http://dl.bintray.com/kotlin/kotlin-eap</url>
+        </repository>
```
skalarproduktraum: Can we move this to non-EAP builds?

pom.xml
```diff
@@ -519,6 +536,21 @@
             <id>clearvolume</id>
             <url>http://dl.bintray.com/clearvolume/ClearVolume</url>
         </repository>
+        <repository>
```
skalarproduktraum: I'm not sure about Jitpack, especially when it comes to reproducible builds. Can you deploy your artifacts to Maven Central e.g.?

src/main/kotlin/graphics/scenery/Settings.kt
```diff
@@ -20,7 +20,7 @@ open class Settings(override var hub: Hub? = null) : Hubable {
      * @param[name] The name of the setting
      * @return The setting as type T
      */
-    fun <T> get(name: String): T {
+    operator fun <T> get(name: String): T {
```
skalarproduktraum: I'd like to keep them as explicit get/set, mainly due to the template parameters necessary sometimes.

src/main/kotlin/graphics/scenery/backends/vulkan/VU.kt
```diff
@@ -38,561 +39,487 @@ fun BigInteger.toHexString(): String {
...
-                val layoutBinding = VkDescriptorSetLayoutBinding.callocStack(descriptorNum, stack)
-                (binding until descriptorNum).forEach { i ->
-                    layoutBinding[i]
-                        .binding(i)
-                        .descriptorType(type)
-                        .descriptorCount(descriptorCount)
-                        .stageFlags(shaderStages)
-                        .pImmutableSamplers(null)
-                }
+    fun createDescriptorSetLayout(
```
skalarproduktraum: Indentation looks like one level too little, could that be?

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanCommandBuffer.kt
```diff
@@ -1,70 +1,48 @@
...
-        val fc = VkFenceCreateInfo.calloc()
-            .sType(VK_STRUCTURE_TYPE_FENCE_CREATE_INFO)
-            .pNext(NULL)
-
-        val f = VU.getLong("Creating fence",
-            { vkCreateFence(device.vulkanDevice, fc, null, this) },
-            { fc.free() })
-
-        fence.put(0, f)
+        fence = device.vulkanDevice createFence vk.FenceCreateInfo { }
```
skalarproduktraum: The use of infix functions in these cases looks unintuitive to me. Would extension functions be an alternative?

pom.xml
elect86: I have been using it extensively for all my projects since quite a while.. only two downtimes in a couple of years

The problem with Maven is that the process is quite an hassle and I have absolutely not experience with that..

pom.xml
elect86: Sure, working on that

src/main/kotlin/graphics/scenery/Settings.kt
elect86: The thing is, you can still use them as explicit.

Although, I think that 

`val value: Int = settings["ciao"]`

is pretty self-explanatory 

Moreover, when the template parameter is necessary, you cannot use the operator

but your project, your rules

src/main/kotlin/graphics/scenery/backends/vulkan/VU.kt
elect86: no idea, I'll double auto format the code

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanCommandBuffer.kt
elect86: you mean 

`device.vulkanDevice.createFence(vk.FenceCreateInfo { })`

?

skalarproduktraum:
Development moved to `VkEnhanced` branch.

PR #198, opened 2018-07-09
Make accumulator, currentTime, t, and shouldClose protected

kephale:
This is necessary when inheriting from SceneryBase *and* overriding the
main() method. All of these are necessary inside any overriding main()
method.


skalarproduktraum:
image

PR #221, opened 2018-11-04
upgrade coremem dependency

haesleinhuepf:
Hi @skalarproduktraum 

in order to integrate SciView / Scenery with Clearcontrol, I'm suggesting upgrading scenerys CoreMem version. I don't see any side effects but I'm also new to Scenery. Please check/test carefully before merging ;-)

Also: I'm just highlighting that groupId and artifactId of coremem changed. That might make things complicated as coremem.CoreMem and net.clearcontrol.coremem can coexist in one project. However both bring the same classes. Thus, further cleaning up the dependency tree might be necessary.

Thanks!

Cheers,
Robert

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -267,9 +267,9 @@
         </dependency>
 
         <dependency>
-            <groupId>net.coremem</groupId>
-            <artifactId>CoreMem</artifactId>
-            <version>0.4.3</version>
+            <groupId>net.clearcontrol</groupId>
+            <artifactId>coremem</artifactId>
+            <version>0.4.6</version>
             <exclusions>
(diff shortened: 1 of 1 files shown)


skalarproduktraum:
Thank you, I shall check this. Should not be too complicated as I am actually only using the data types and so on. Might be worth looking into expanding scenery to support loading data directly from CoreMem objects.

haesleinhuepf:
Alternatively, can one import OpenCL objects into Scenery?

skalarproduktraum:
Not yet, but that should not be too difficult to implement. Could you open an issue for that please?

PR #258, opened 2019-07-22
Add betweenPoints method for Cone

kephale:


skalarproduktraum:
Hey Kyle,

why do you need this routine? There is already Node.orientBetweenPoints(), is this insufficient?

kephale:
Because it constructs the cone at the same time, like `Cylinder.betweenPoints`

skalarproduktraum:
Okay, sounds good to me 👍 Could you please add docs for the function, and maybe a test? Then we're good 😎

PR #259, opened 2019-08-22
Simplify native profiles

ctrueden:
Starting with pom-scijava-base 7.0.0, some native-related properties are now defined automatically. See:

https://github.com/scijava/pom-scijava-base/commit/42759cc96b81bb0cdd2bb5e1fa23567c77962f2e

The pom-scijava 27 series extends a new-enough pom-scijava-base to inherit these properties. So we can reuse them here without needing to define our own explicit profiles anymore.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
     <parent>
         <groupId>org.scijava</groupId>
         <artifactId>pom-scijava</artifactId>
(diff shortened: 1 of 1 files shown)


skalarproduktraum:
Sorry, @ctrueden, I had missed your PR! Thank you very much, anything that simplifies stuff is great 👍

PR #262, opened 2019-10-07
Log name of failed searches during native extraction

kephale:


skalarproduktraum:
This is a nice and useful change, thanks!

PR #264, opened 2019-10-08
Override toString for Node

kephale:


skalarproduktraum:
This is pretty nice, I would suggest not using `nodeType` though, because this field should actually have been deprecated a while ago, and is not used anywhere. How about class.simpleName or sth like that instead?

PR #270, opened 2019-10-30
Fix material usage in PointCloud and PointCloudExample

kephale:


src/test/tests/graphics/scenery/tests/examples/basic/PointCloudExample.kt
```diff
@@ -18,6 +18,7 @@ class PointCloudExample : SceneryBase("PointCloudExample") {
 
     override fun init() {
         renderer = hub.add(Renderer.createRenderer(hub, applicationName, scene, windowWidth, windowHeight))
+        renderer!!.pushMode = true
```
skalarproduktraum: Don't use .!! unless absolutely necessary, use .? instead 👍

src/main/kotlin/graphics/scenery/PointCloud.kt
```diff
@@ -36,6 +36,14 @@ open class PointCloud(var pointRadius: Float = 0.1f, override var name: String =
         // we are going to use shader files whose name is derived from the class name.
         // -> PointCloud.vert, PointCloud.frag
         material = ShaderMaterial.fromClass(this::class.java)
+        material.cullingMode = Material.CullingMode.None
+        material.blending.transparent = true
```
skalarproduktraum: Are PointClouds always supposed to be transparent?

PR #291, opened 2020-02-14
Rendering device chooser

xulman:
The `VulkanRenderer` backend now saves names of all vulkan devices it has iterated over when it was starting/created/initiated. The list is saved in `VulkanRenderer.discoveredDevices` array and is accessible to the outer world (and used inside sciview menu item that allows user to choose different renderer).

@skalarproduktraum please polish the code, I'm not a Kotlin guru and always forgetting language basics.


kephale:
Note that this needs to be merged before the downstream sciview menu option can be merged

xulman:
sorry, I've accidentally closed this PR (while the downstream one at sciview https://github.com/scenerygraphics/sciview/pull/255  is still there) when I was cleaning branches in my fork...

ctrueden:
@xulman @kephale @skalarproduktraum I rebased this against the current mainline branch.

PR #294, opened 2020-02-28
Curve

Domino2357:
Contains a class for Catmull Rom Splines, a geometry which can be rendered around those splines, and a small example.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -135,7 +135,7 @@
         <reflections.version>0.9.11</reflections.version>
         <script-editor.version>0.5.3</script-editor.version>
         <slf4j.version>1.7.25</slf4j.version>
-        <spirvcrossj.version>0.6.0-1.1.106.0</spirvcrossj.version>
+        <spirvcrossj.version>0.7.0-1.1.106.0</spirvcrossj.version>
         <trove4j.version>3.0.3</trove4j.version>
 
         <lwjgl.natives>natives-${scijava.platform.family.long}</lwjgl.natives>
--- a/pom.xml
+++ b/pom.xml
@@ -541,6 +541,7 @@
             <artifactId>lwjgl3-awt</artifactId>
             <version>${lwjgl3-awt.version}</version>
         </dependency>
+
     </dependencies>
 
     <repositories>
--- a/src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
+++ b/src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
@@ -39,6 +39,7 @@ class CatmullRomSpline(val controlPoints: List<GLVector>, val alpha: Float = 0.5
 
         var t = t1
         while(t<t2) {
+            //The t's must not be equal, otherwise we divide by zero
             if(t1 != t0 && t2 != t1 && t2 != t0 && t3 != t1 && t3 != t2) {
                 val A1 = p0.times((t1 - t) / (t1 - t0)) + p1.times((t - t0) / (t1 - t0));
                 val A2 = p1.times((t2 - t) / (t2 - t1)) + p2.times((t - t1) / (t2 - t1));
--- a/src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
+++ b/src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
@@ -39,17 +39,23 @@ class CatmullRomSpline(val controlPoints: List<GLVector>, val alpha: Float = 0.5
 
         var t = t1
         while(t<t2) {
-            val A1 = p0.times((t1-t)/(t1-t0)) + p1.times((t-t0)/(t1-t0));
-            val A2 = p1.times((t2-t)/(t2-t1)) + p2.times((t-t1)/(t2-t1));
-            val A3 = p2.times((t3-t)/(t3-t2)) + p3.times((t-t2)/(t3-t2));
+            if(t1 != t0 && t2 != t1 && t2 != t0 && t3 != t1 && t3 != t2) {
+                val A1 = p0.times((t1 - t) / (t1 - t0)) + p1.times((t - t0) / (t1 - t0));
+                val A2 = p1.times((t2 - t) / (t2 - t1)) + p2.times((t - t1) / (t2 - t1));
+                val A3 = p2.times((t3 - t) / (t3 - t2)) + p3.times((t - t2) / (t3 - t2));
 
-            val B1 = A1.times((t2-t)/(t2-t0)) + A2.times((t-t0)/(t2-t0));
-            val B2 = A2.times((t3-t)/(t3-t1)) + A3.times((t-t1)/(t3-t1));
+                val B1 = A1.times((t2 - t) / (t2 - t0)) + A2.times((t - t0) / (t2 - t0));
+                val B2 = A2.times((t3 - t) / (t3 - t1)) + A3.times((t - t1) / (t3 - t1));
 
-            val C = B1.times((t2-t)/(t2-t1)) + B2.times((t-t1)/(t2-t1));
-            curvePoints.add(C)
+                val C = B1.times((t2 - t) / (t2 - t1)) + B2.times((t - t1) / (t2 - t1));
+                curvePoints.add(C)
 
-            t += ((t2-t1)/n)
+                t += ((t2 - t1) / n)
+            }
+            else {
+                throw IllegalArgumentException("The intermediate products of the calculations must not be equal!" +
+                    "Otherwise we devide by zero.")
+            }
         }
 
         return curvePoints
@@ -61,7 +67,6 @@ class CatmullRomSpline(val controlPoints: List<GLVector>, val alpha: Float = 0.5
      */
     fun catMullRomChain(n: Int = 100): ArrayList<GLVector> {
         val chainPoints = ArrayList<GLVector>()
-        val j = controlPoints.size-4
         controlPoints.dropLast(3).forEachIndexed {  index, _ ->
             val c = CatmulRomSpline(controlPoints[index], controlPoints[index+1],
                 controlPoints[index+2], controlPoints[index+3], n)
@@ -71,4 +76,3 @@ class CatmullRomSpline(val controlPoints: List<GLVector>, val alpha: Float = 0.5
     }
 
 }
-
--- a/src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
+++ b/src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
@@ -41,15 +41,15 @@ class CatmullRomSpline(val controlPoints: List<GLVector>, val alpha: Float = 0.5
         while(t<t2) {
             //The t's must not be equal, otherwise we divide by zero
             if(t1 != t0 && t2 != t1 && t2 != t0 && t3 != t1 && t3 != t2) {
-                val A1 = p0.times((t1 - t) / (t1 - t0)) + p1.times((t - t0) / (t1 - t0));
-                val A2 = p1.times((t2 - t) / (t2 - t1)) + p2.times((t - t1) / (t2 - t1));
-                val A3 = p2.times((t3 - t) / (t3 - t2)) + p3.times((t - t2) / (t3 - t2));
+                val a1 = p0.times((t1 - t) / (t1 - t0)) + p1.times((t - t0) / (t1 - t0))
+                val a2 = p1.times((t2 - t) / (t2 - t1)) + p2.times((t - t1) / (t2 - t1))
+                val a3 = p2.times((t3 - t) / (t3 - t2)) + p3.times((t - t2) / (t3 - t2))
 
-                val B1 = A1.times((t2 - t) / (t2 - t0)) + A2.times((t - t0) / (t2 - t0));
-                val B2 = A2.times((t3 - t) / (t3 - t1)) + A3.times((t - t1) / (t3 - t1));
+                val b1 = a1.times((t2 - t) / (t2 - t0)) + a2.times((t - t0) / (t2 - t0))
+                val b2 = a2.times((t3 - t) / (t3 - t1)) + a3.times((t - t1) / (t3 - t1))
 
-                val C = B1.times((t2 - t) / (t2 - t1)) + B2.times((t - t1) / (t2 - t1));
-                curvePoints.add(C)
+                val c = b1.times((t2 - t) / (t2 - t1)) + b2.times((t - t1) / (t2 - t1))
+                curvePoints.add(c)
 
                 t += ((t2 - t1) / n)
             }
--- a/src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
+++ b/src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
@@ -1,0 +1,74 @@
+package graphics.scenery
+
+import cleargl.GLVector
+import kotlin.math.pow
+
+/**
+ * This class offers the logic for creating a CatmuLL Rom Spline. This is essentially
+ * a spline going through control points. For more information see:
+ * https://en.wikipedia.org/wiki/Centripetal_Catmull–Rom_spline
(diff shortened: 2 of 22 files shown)


pom.xml
```diff
@@ -541,6 +541,19 @@
...
+            <artifactId>biojava-core</artifactId>
+            <version>5.3.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.biojava</groupId>
+            <artifactId>biojava-structure</artifactId>
+            <version>5.3.0</version>
+        </dependency>
+
```
skalarproduktraum: I think the biojava dependencies are not (yet) relevant ;) Please remove!

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
```diff
@@ -0,0 +1,71 @@
...
+
+import cleargl.GLVector
+import kotlin.math.pow
+
+/**
+ * This class offers the logic for creating a CatmuLL Rom Spline. This is essentially
+ * a spline going through control points. For more information see:
+ * https://en.wikipedia.org/wiki/Centripetal_Catmull–Rom_spline
+ * @param controlPoints the list of control points
+ * @param alpha determines the kind of Catmull Rom Spline, set in range of 0..1
```
skalarproduktraum: This is a bit unclear, what exactly does the parameter control, and why is the default 0.5?

skalarproduktraum: Also, in kdoc (as opposed to Javadoc), parameters can be referred to by their name via `[myParameter]`, and don't need the `@param` annotation.

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
```diff
@@ -0,0 +1,71 @@
...
+ */
+class CatmullRomSpline(val controlPoints: List<GLVector>, val alpha: Float = 0.5f) {
+    
+    /**
+     * Calculates the parameter t.
```
skalarproduktraum: Okay, but what is t? And what are the parameters?

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
```diff
@@ -0,0 +1,71 @@
...
+        return(((Pj.x()-Pi.x()).pow(2) + (Pj.y()-Pi.y()).pow(2)
+                + (Pj.z()-Pi.z()).pow(2)).pow(exp) + ti)
+    }
+
+    /**
+     * this function returns the spline points between two points. Please note you need four points
+     * to have a smooth curve.
+     * @param n number of points defining the spline
+     */
+    fun CatmulRomSpline(P0: GLVector, P1: GLVector, P2: GLVector, P3: GLVector, n: Int = 100): List<GLVector> {
```
skalarproduktraum: Parameter names should be lowercase :) Is n the number of control points? I would then expect it to appear in the allocation for curvePoints. It however only appears as normalisation factor for t.

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
```diff
@@ -0,0 +1,71 @@
...
+        val j = controlPoints.size-4
+        for (i in 0..j) {
+            val c = CatmulRomSpline(controlPoints[i], controlPoints[i+1],
+                    controlPoints[i+2], controlPoints[i+3], n)
+            chainPoints.addAll(c)
+        }
+        return chainPoints
+    }
+
+}
```
skalarproduktraum: Please add a newline at the end of the file!

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    /**
+     * Returns the actual curve with all the points.
+     * @param n number of points the curve has
+     */
+    fun CatMulRomChain(n: Int = 100): ArrayList<GLVector> {
+        val chainPoints = ArrayList<GLVector>()
+        val j = controlPoints.size-4
+        for (i in 0..j) {
```
skalarproduktraum: Is this intended to run over all i (instead of `i in 0..j step 4`)?

src/main/kotlin/graphics/scenery/CurveGeometry.kt
```diff
@@ -0,0 +1,189 @@
+package graphics.scenery.proteins
+
+import cleargl.GLMatrix
+import cleargl.GLVector
+import com.jogamp.opengl.math.FloatUtil.makeRotationAxis
+import graphics.scenery.*
+import org.lwjgl.opengl.GL
```
skalarproduktraum: Why is this import here?

src/main/kotlin/graphics/scenery/CurveGeometry.kt
```diff
@@ -0,0 +1,189 @@
...
+    private val cur = curve.CatMulRomChain(n)
+
+    /**
+     * This function renders the spline.
+     * @param [baseShape] It takes a lambda as a parameter, which is the shape of the
+     * curve. If you choose, for example, to have a square as a base shape, your spline will look like
+     * a banister. Please not that the base shape needs an equal number of points in each segments but it
+     * can very well vary in thickness.
+     */
+    fun drawSpline(baseShape: (() -> List<GLVector>)){
```
skalarproduktraum: Space between ) and {

src/main/kotlin/graphics/scenery/CurveGeometry.kt
```diff
@@ -0,0 +1,189 @@
...
+        data class TranslationMatrix(val matrix: GLMatrix, val translation: GLVector)
+        val bases = ArrayList<TranslationMatrix>()
```
skalarproduktraum: Make sure you pre-allocate the correct ArrayList sizes everywhere (if you can calculate them, otherwise estimate!) -- otherwise, the backing array of the list will need to be reallocated again and again when you add stuff.

src/main/kotlin/graphics/scenery/CurveGeometry.kt
```diff
@@ -0,0 +1,189 @@
...
+                basisArray.add(t.x())
+                basisArray.add(0f)
+                basisArray.add(n.y())
+                basisArray.add(b.y())
+                basisArray.add(t.y())
+                basisArray.add(0f)
+                basisArray.add(n.z())
+                basisArray.add(b.z())
+                basisArray.add(t.z())
+                basisArray.add(0f)
```
skalarproduktraum: Instead of adding zeros here, you can stuff in the vector, making the additional GLVector superfluous (this is actually also how translations with homogeneous matrices work)

src/main/kotlin/graphics/scenery/CurveGeometry.kt
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        val curveGeometry = ArrayList<ArrayList<GLVector>>()
+        bases.forEach { (m,t) ->
+            val basis = m.inverse
+            val shape = ArrayList<GLVector>()
+            baseShape.invoke().forEach {
+                val vector4D = GLVector(it.x(), it.y(), it.z(), 0f)
+                val vector = basis.mult(vector4D)
+                val vertex = GLVector(vector.x()+t.x(), vector.y()+t.y(), vector.z()+t.z())
```
skalarproduktraum: See above, putting the vector in the 4x4 matrix would rid you of the addition here, because the matrix multiplication would handle it already.

src/main/kotlin/graphics/scenery/CurveGeometry.kt
```diff
@@ -0,0 +1,189 @@
...
+            }
+            curveGeometry.add(shape)
+        }
+
+        val verticesVectors = ArrayList<GLVector>()
+
+        curveGeometry.forEachIndexed { shapeIndex, shape ->
+            if (shapeIndex < curveGeometry.lastIndex) {
+                shape.forEachIndexed { vertexIndex, vertex ->
+                    if(vertexIndex < shape.lastIndex) {
```
skalarproduktraum: Instead of having an if for each of the indices, consider doing sth like `shape.dropLast(1).forEachIndexed`

src/main/kotlin/graphics/scenery/CurveGeometry.kt
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+
+    /**
+     * This function returns the frenet frames along the curve. This is essentially a new
+     * coordinate system which represents the form of the curve. For details concerning the
+     * calculation see: http://www.cs.indiana.edu/pub/techreports/TR425.pdf
+     */
+    data class FrenetFrame(val tangent: GLVector, var normal: GLVector?, var bitangent: GLVector?, val translation: GLVector)
+    fun computeFrenetFrames(): List<FrenetFrame> {
```
skalarproduktraum: Missing kdoc for computeFrenetFrames()

src/main/kotlin/graphics/scenery/CurveGeometry.kt
```diff
@@ -0,0 +1,189 @@
...
+        else {
+            GLVector(1f, 0f, 0f)
+        }
+
+        val normal = frenetFrameList[0].tangent.cross(vec)
+
+        frenetFrameList[0].normal = normal
+        frenetFrameList[0].bitangent = frenetFrameList[0].tangent.cross(normal)
+
+        for(i in 0 until (frenetFrameList.size-1)) {
```
skalarproduktraum: Instead of indexing, use `frenetFrameList.windowed(2, 1).forEach { (firstFrame, secondFrame) -> ...`

src/main/kotlin/graphics/scenery/volumes/AminoAcidBonds.kt
```diff
@@ -0,0 +1,2 @@
+package graphics.scenery.volumes
```
skalarproduktraum: This should not be here at all :-D And `graphics.scenery.volumes` is also the totally wrong package for it.

src/main/kotlin/graphics/scenery/CurveGeometry.kt
```diff
@@ -0,0 +1,189 @@
+package graphics.scenery.proteins
```
skalarproduktraum: This should live in the `graphics.scenery` package.

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
```diff
@@ -0,0 +1,71 @@
+package graphics.scenery.proteins
```
skalarproduktraum: This should live in the `graphics.scenery` package.

src/test/tests/graphics/scenery/tests/examples/advanced/CurveGeometryExample.kt
```diff
@@ -0,0 +1,102 @@
+package graphics.scenery.tests.examples.PDB
```
skalarproduktraum: This should live in the `graphics.scenery.tests.examples.basic` (or advanced, your call) package.

src/test/tests/graphics/scenery/tests/examples/advanced/CurveGeometryExample.kt
```diff
@@ -0,0 +1,102 @@
...
+    override fun inputSetup() {
+        super.inputSetup()
+        setupCameraModeSwitching()
+    }
+
+    @Test
+    override fun main() {
+        super.main()
+    }
+}
```
skalarproduktraum: New line missing.

pom.xml
skalarproduktraum: These guys are also the reason for the CI builds failing atm.

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
Domino2357: Nice to know! Well, my description was incorrect: n is in fact the number of points generated between the two control points, e.g. n = 3, then the generated points between (0,0,0) and (1,0,0) would be (0.25,0,0), (0.5,0,0), and (0.75,0,0).

src/main/kotlin/graphics/scenery/volumes/AminoAcidBonds.kt
Domino2357: Oh, that wasn't on purpose. :D Good thing you saw it!

src/main/kotlin/graphics/scenery/CurveGeometry.kt
Domino2357: Hm, I remember now why it was so laborious: because the matrix needed to be inversed to give the baseShape-points their new coordinate system. However, the translation must not be inverted. I tried now to invert the matrix and then use matrix.translate() to do the translation but it doesn't work that way. Do you have any ideas?

src/test/tests/graphics/scenery/tests/unit/CurveGeometryTests.kt
```diff
@@ -13,58 +13,58 @@ import kotlin.test.assertNotNull
...
-        assertEquals(frenetFrames.filter { it.bitangent?.length2()!! < 1.1f && it.bitangent?.length2()!! > 0.9f },
+        assertEquals(frenetFrames.filter { it.bitangent?.length2()!! < 1.0001f && it.bitangent?.length2()!! > 0.99999f },
                 frenetFrames)
-        assertEquals(frenetFrames.filter { it.normal?.length2()!! < 1.1f && it.normal?.length2()!! > 0.9f },
+        assertEquals(frenetFrames.filter { it.normal?.length2()!! < 1.0001f && it.normal?.length2()!! > 0.99999f },
                 frenetFrames)
-        assertEquals(frenetFrames.filter { it.tangent.length2() < 1.1f && it.tangent.length2() > 0.9f },
+        assertEquals(frenetFrames.filter { it.tangent.length2() < 1.0001f && it.tangent.length2() > 0.99999f },
                 frenetFrames)
         assertFails {  geometry.drawSpline { triangleFalse() } }
```
skalarproduktraum: This test actually tests two things: Correct creation of Frenet frames, and failure of creation in case the number of vertices is not constant. Could you please split that into two tests that reflect those facts in their name? Thanks!

src/test/tests/graphics/scenery/tests/unit/CatmullRomSplineTests.kt
```diff
@@ -16,29 +16,23 @@ import graphics.scenery.numerics.Random
...
 
-        val controlPoints = arrayListOf(GLVector(x1, y1, z1), GLVector(x2, y2, z2), GLVector(x3, y3, z3),
-        GLVector(x4, y4, z4))
+        val controlPoints = arrayListOf(point1, point2, point3, point4)
 
         val curve = CatmullRomSpline(controlPoints)
-
         assertNotNull(curve)
-        assertTrue(curve.catMullRomChain().size == 100)
+        //The computation of the Catmull Rom Spline can deliver sometimes one additional point
```
skalarproduktraum: Uh... sometimes? Could you please figure out why that is? The algorithm contains no random parts, so it should be entirely deterministic.

Domino2357: My formulation was not good- the difference occurs because of the random vectors. It is a bit abstract, but in the end one gets 101 points if the distance between point 0 and point 1 is small relative to the difference between point 1 and point 2. In that case the while loop will spit out one more point.
However, it can't be more than 101 or less than 100. I will update the comment of course. What do you think, can it lead to problems?

src/test/tests/graphics/scenery/tests/unit/CurveGeometryTests.kt
Domino2357: Sure!

Domino2357: Oh, do you mean two different classes or two different functions?

PR #298, opened 2020-03-19
Curve Geometry

Domino2357:
This pull request contains the catmull rom spline and a the geometry riding along the curve as well as an example and some unit test.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -541,6 +541,7 @@
             <artifactId>lwjgl3-awt</artifactId>
             <version>${lwjgl3-awt.version}</version>
         </dependency>
+
     </dependencies>
 
     <repositories>
--- a/pom.xml
+++ b/pom.xml
@@ -541,7 +541,6 @@
             <artifactId>lwjgl3-awt</artifactId>
             <version>${lwjgl3-awt.version}</version>
         </dependency>
-
     </dependencies>
 
     <repositories>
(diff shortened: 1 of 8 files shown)


Domino2357:
Heyho!
I have no idea why this doesn't work with the other pull request but this new pull request contains everything we want without the redundant commits. Would it be a problem to discard the old pull request and use this one instead, @skalarproduktraum ?

skalarproduktraum:
No problem to close the old one and use this one, we just need to see why the CI is now failing ;)

skalarproduktraum:
So there still seems to be an issue with your unit tests, please verify. The failures on both Travis and Appveyor are genuine failures. Please also try to run the tests locally via `mvn test`.

Domino2357:
@skalarproduktraum Finally! :)

skalarproduktraum:
Great! Four (hopefully :-D) final things:

* why is the `drawSpline` method there? Couldn't that stuff already be done in the constructor? Feels like a bit of an ugly API, sorry
* please rename `CurveGeometry` to `Curve`
* in CatmullRomSplineTests, the test is a little bit weak, just testing that the curve creation doesn't fail. Could you come up with something more exhaustive there please?
* pom.xml got an extra new line that should not be there

Domino2357:
@skalarproduktraum I hope this is it. If you still have improvements I could do, please let me know!

skalarproduktraum:
Looks good to me, thank you!

PR #299, opened 2020-03-27
Spline

Domino2357:
Contains the logic of the splines (Spline is a abstract class) as well as the UniformBSpline-class with an example and an unit test. I changed the CatmullRomSpline, it does now inherit from Spline. The Curve now renders a geometry along any Spline.


skalarproduktraum:
Hey @Domino2357! Thanks for the PR! Now that we've merged the move to JOML's Vector*f/Matrix4f, this PR will unfortunately need some work from you. Please let me know if you need help!

skalarproduktraum:
Hey @Domino2357! I think there is something wrong here -- all those commits from master and bvv-updates-sources should not be in this PR at all. Could you please put your changes in a new branch based on the current master and recreate this PR? This will be a terrible mess otherwise. Thanks!

Domino2357:
Heyho @skalarproduktraum ! So the code is now running and showing in the examples what it is supposed to show. Did you clean up my commits? Because I can't see the commits from the master anymore.
About the test: I ran each test-class at least ten times as well as `mvn test` (also ten times) without failure. Also the appveyor works fine. Do you have an error message? Else, I would search for a script, which just runs the test over and over until something breaks.

skalarproduktraum:
Hey @Domino2357! I actually did not clean up anything, I just merged another branch yesterday, maybe you were on this one for some reason. But this looks way better now. 

Regarding the unit tests, have a look at this run for example: https://ci.appveyor.com/project/skalarproduktraum/scenery/builds/32661266

Domino2357:
Alright! Thanks to your trace, I now know what the problem was: the test verified if the points of the spline were equidistant. The tolerance was apparently too fine tuned. I increased the tolerance tenfold, given that the test passed so many times, this should be sufficient.

skalarproduktraum:
Great you found it! Could you please add some comments in the unit tests though, why certain values are chosen as they are? For me, it's totally opaque why this value is 1.0 and was 0.5 before.

Domino2357:
@skalarproduktraum Thanks for poking a little bit! The value was arbitrary I have to admit. But I thought about it and it almost stayed the same, but now it has an explanation.

Domino2357:
Oh, and all those strange commits were still there as untracked files. But since I didn't commit them they didn't get pushed obviously, just wanted to let you know.

src/main/kotlin/graphics/scenery/Spline.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.joml.Vector3f
+
+/**
+ * This is the abstract class for all the spline classes currently existing. It takes a list
+ * of [controlPoints], which constrict spline. Moreover, one has to specify the number of points
+ * in one segment of the spline [n].
+ *
+ * @author Justin Bürger <burger@mpi-cbg.de>
+ */
+abstract class Spline(open val controlPoints: List<Vector3f>, open val n: Int) {
```
skalarproduktraum, line 12: Considering this doesn't really need to store state, maybe it would make more sense to make it an interface instead of an abstract class?

src/main/kotlin/graphics/scenery/Spline.kt
```diff
@@ -0,0 +1,17 @@
...
+    /**
+     * Returns all the points the spline has.
+     */
+    abstract fun splinePoints(): ArrayList<Vector3f>
```
skalarproduktraum, line 16: List<> is probably enough here

skalarproduktraum:
Hey @Domino2357! Just two minor points, please have a look. I am also fine with Spline being an abstract class, if necessary. But consider that inheritance on the JVM only allows inheriting from a single class, but from multiple interfaces. So if you'd have a DoubleHelix class for example, implementing Spline, it'd actually also need to inherit from Mesh. So I guess it should rather be an Interface.

skalarproduktraum:
Awesome! One more thing: Do you want to squash some of the commits together, or shall I just merge them all as a single commit?

Domino2357:
@skalarproduktraum Great! One commit will do it, there is no extra contribution there worth its own commit.

skalarproduktraum:
Done, thank you very much!

PR #301, opened 2020-04-04
Fix maximum bounding box for Nodes with children

kephale:
The position of the children is defined w.r.t. the parent and must be used as an offset when finding the maximum bounding box.


skalarproduktraum:
Thank you, @kephale!

PR #319, opened 2020-06-09
POM: remove redundant version tags

imagejan:
* remove all version tags for artifacts managed by `pom-scijava`
* remove hard-coded version pinning for `n5` artifacts
* override `jacoco-maven-plugin.version` property instead of hard-coding the version
* remove redundant `scijava.jvm.version` specified in parent POM


skalarproduktraum:
Thanks, @imagejan!

PR #320, opened 2020-06-12
camera moves no longer bound to y-axis

xulman:
Hi,
the original rotation implemented in both ArcBall and FPS camera moves is a composition of two moves: vertical along view x-axis and horizontal along y-axis. Original code, however had the latter acting upon y-axis that got tilted as part of the first vertical move. The proposed code does independent x- and y- axes rotation.

The dependecy/tilting of y-axis rendered cam moves schizofrenic:
- when object is "up front" - no vertical (pitch) cam move, the horizontal mouse move showed the expected horizontal (yaw) move of the cam/object
- when object is 90deg "pitched", the horizontal mouse move rendered into *rotation* along the forward-view axis (the tilted y)

Some views were impossible to achive.. .this is perhaps the glimball lock (mispeled most likely)

(I'm now working on more UX commits on the SciView side to complement this, stay tuned :-)
Vlado


xulman:
I will rebase this whole PR (and drop this one and create new oneS), please ignore it for now (and sorry for that)

xulman:
This PR is now OBSOLETE. Closing it.

PR #323, opened 2020-07-08
Cam controls add-ons

xulman:
A few tiny changes towards better and nicer scenery (and SciView) :-)

### DIFF
--- a/src/main/kotlin/graphics/scenery/controls/behaviours/ArcballCameraControl.kt
+++ b/src/main/kotlin/graphics/scenery/controls/behaviours/ArcballCameraControl.kt
@@ -15,14 +15,15 @@ import kotlin.reflect.KProperty
  * Targeted ArcBall control
  *
  * This Behaviour provides ArcBall control for scenery, with a customizable target. If you
- * activate this behaviour, it'll use the current camera distance to the target as initial distance.
+ * activate this behaviour or switch target, it'll be operating the camera from the current
+ * distance to the current target.
  *
  * The Targeted ArcBall also provides [minimumDistance] and [maximumDistance] to clamp the distance
  * to the target to this range.
  *
  * @author Ulrik Günther <hello@ulrik.is>
  * @property[name] The name of the behaviour
- * @property[node] The node this behaviour controls
+ * @property[camNode] The node this behaviour controls
  * @property[w] Window width
  * @property[h] Window height
  * @property[target] [Vector3f]-supplying with the look-at target of the arcball
@@ -34,7 +35,7 @@ open class ArcballCameraControl(private val name: String, private val n: () -> C
     private var firstEntered = true
 
     /** The [graphics.scenery.Node] this behaviour class controls */
-    protected var node: Camera? by CameraDelegate()
+    protected var camNode: Camera? by CameraDelegate()
 
     /** Camera delegate class, converting lambdas to Cameras. */
     protected inner class CameraDelegate {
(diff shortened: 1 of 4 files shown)


xulman:
That commit introduces explicit "mouse move sensitivity" attribute into FPS -- it is implemented entirely ala Arcball

xulman:
(hmm, was expecting commits listing and messages to be interleaved)

The arcball-commits:
- code tidy
- fix when scrolling/zooming-in-out: initial `distance` was set a node was selected, it could have been moved since then but the `distance` got no update, then zooming-in-out was not doing the expected job (because it had no accurate `distance` info)

xulman:
(also the order of commits is changed, interesting...: the MovementCOmmand-commit should not be between the two for arcball...)

the `MovementCommand` is code cosmetics and a bit of optimization: rather then computing `cross()` freshly, we leverage on the camera's `right()` vector (which is always updated with every camera change and so here the `cross()` is not computed 2nd-time)

skalarproduktraum:
So this PR actually introduces something I did deliberately prevent: rotation around the axis of sight. What's the reasoning behind introducing this? If truly necessary, I'd propose creating an additional CameraControl on top of ArcballCameraControl to include this.

xulman:
ok, please then ignore the last commit (https://github.com/scenerygraphics/scenery/pull/323/commits/6cf9fb5a6ccf6c6bc6f0ec8e9bfdb6682cd9581e) in this PR and consider only the rest.. so that we could close it potentially

xulman:
sooo:
- I've reverted (via the force push that replaced only the last commit 6cf9fb5) any change to the original ArcBall and FPS cam controls -- they should stay intact in this PR functionality-wise,
- I would love to have added the new class (class name is welcomed subject to change) that does ArcBall with the sticky-y-axis (formely branded as flexi-y-axis)
- aaah, actually, I propose to rename `node` to `camNode` in ArcBall and updated doc (in a hope that difference between the two arcballs would be clear(er)

I couldn't see easy way to implant the *use of* the new class into scenery, we would have to change
https://github.com/scenerygraphics/scenery/blob/cefd4470ef7f1fdc055c55ad82cdce07f9933083/src/main/kotlin/graphics/scenery/SceneryBase.kt#L376
most likely and this one has state variable, and... 

I wanted to have the stickyYArcball as a *regime* of the original, so either you work with FPS or with ArcBall (and you could separately define which particular ArcBall should be used)... so it shall not be FPS->ArcBall->StickyYArcBall->FPS but rather FPS->oneOfArcBall->FPS

As a result, the new class in not used anywhere in scenery.... (sorry for that). But, I use it in sciview:
https://github.com/scenerygraphics/sciview/compare/master...xulman:addingStickyYArcball
where you can ctrl-shift-A-Y  (even a cat would have a hard time to turn it on by accident) replace what is used for Arcballing

Would that be good?

-----------------
I was wondering about a boolean switch in the original Acrball.kt... I think it would have made everything much simpler, less error-prone in sciview code (there would have been always on object for arcball, not two like I am proposing above, no need to update always both to have them synced), but such flag is most likely not clean code type of thing (on the other hand, one can think of it as of another param to the code, similarily to, e.g., mouseSpeed param -- we also don't have separate arcball object for every speed rate)

I am saying if possibly the https://github.com/scenerygraphics/sciview/pull/273 becomes a reality, we would need to extend the https://github.com/scenerygraphics/sciview/pull/273/commits/871d12e7604a6354aecd9bf0ae7c5a6f645556c9 - a governing class that updates/forwards controls param changes to the relevant controllers, and most likely more I don't even *remember now*... (that is the motor for me to be writing this paragraphs)

xulman:
deprecated in favour of #329

PR #324, opened 2020-07-08
inputHandler add ons

xulman:
here's a couple of things I would really use to have for SciView:

- the `InputHandler.openKeybindingsGuiEditor()` that opens (also in scenery, in theory) the key bindings GUI editor
- the `getAllBehaviours()` listing/overview method and `getKeyBindings()` query method
- also `removeKeyBinding()` (and any binding updating method in `INputHandler`) are maintaining the synchronization of ui-behaviours `InputTriggerMapping` and `InputTriggerConfig` explicitly and on its own (....and are now functional)
- these items need the PRs from ui-behaviour... this PR will not compile without it

Plus, I noticed overlapping key bindings:
- gamepad_cam_control was using 'P' just like screenshots, I remapped that to 'G' (for *G*amepad)
- but 'G' was used for mouse_control, so I remapped it to 'M' (for *M*ouse)

Also noticed that scenery's "select_command" gets no function to process the result of ray-tracing,
so I disabled/commented-out that behaviour.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -110,6 +110,8 @@
         <dokka.version>0.10.1</dokka.version>
         <dokka.skip>true</dokka.skip>
 
+        <ui-behaviour.version>2.0.2</ui-behaviour.version>
+
         <bigvolumeviewer.version>0.1.8</bigvolumeviewer.version>
         <ffmpeg.version>4.2.1-1.5.2</ffmpeg.version>
         <jackson-dataformat-msgpack.version>0.8.20</jackson-dataformat-msgpack.version>
--- a/src/main/kotlin/graphics/scenery/controls/InputHandler.kt
+++ b/src/main/kotlin/graphics/scenery/controls/InputHandler.kt
@@ -285,9 +285,10 @@ class InputHandler(scene: Scene, renderer: Renderer, override var hub: Hub?, for
         frame.isVisible = true
 
         //process "Apply" button of the editor
-        editorPanel.addConfigCommittedListener {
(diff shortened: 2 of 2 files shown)


xulman:
the last commit could/should be *fixup*'ed with https://github.com/scenerygraphics/scenery/pull/324/commits/db207f1935ce469c2e95996f7c66fc5869cc77af

I had to adapt to https://github.com/scijava/ui-behaviour/pull/21#issuecomment-659233027

xulman:
the previous commit could/should be *fixup*'ed with https://github.com/scenerygraphics/scenery/pull/324/commits/8692906677fde3e30ed9b4ba182ce4adb1b2c412

I fixed after advice how to correctly use ui-behaviour: https://github.com/scijava/ui-behaviour/pull/20/files/f99fb9f12b22e4bc9a2d507b882f0c29533d0b7f#r455717091

after https://github.com/scijava/ui-behaviour/pull/20 is merged, and new version of ui-behaviour is released, then this PR is ready to go...

xulman:
the previous commit could/should be *fixup*'ed with https://github.com/scenerygraphics/scenery/pull/324/commits/8692906677fde3e30ed9b4ba182ce4adb1b2c412

I fixed in response to how ui-behaviour#20 ended up

xulman:
finally, the commit updating the pom.xml after which it should finally compile

...but, ups, it is not compiling because, looking into Travis, I violated the rule of not-using-snapshot versions;
hmm, then we have to wait until this: https://github.com/scijava/pom-scijava/pull/138
(and update parent pom version later)

PR #338, opened 2020-10-16
Ribbon

Domino2357:
- provides a visualization of a ribbon diagram
- furthermore, allows different base shapes in a curve


pom.xml
```diff
@@ -563,6 +563,23 @@
             <groupId>org.janelia.saalfeldlab</groupId>
             <artifactId>n5-imglib2</artifactId>
         </dependency>
+        <dependency>
```
skalarproduktraum, line 566: Please have a look at the CI logs -- the new dependencies seem to include more slf4j bindings, leading to duplicate classes. They can be excluded from a specific dependency using the procedure described at https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html#dependency-exclusions

src/main/kotlin/graphics/scenery/Axis.kt
```diff
@@ -0,0 +1,127 @@
+package graphics.scenery
+
+import com.jogamp.opengl.math.FloatUtil
```
skalarproduktraum, line 3: Please don't use JOGL classes.

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
```diff
@@ -14,24 +15,27 @@ import kotlin.math.pow
  * [alpha] determines the kind of Catmull Rom Spline, set in range of 0..1, the
  * resulting curve for alpha = 0 is a standart Catmull Rom Spline, for alpha = 1 we get
  * a chordal Catmull Rom Spline.
+ *
+ * @author  Justin Buerger <burger@mpi-cbg.de>
  */
-class CatmullRomSpline(protected val controlPoints: List<Vector3f>, val n: Int = 100, val alpha: Float = 0.5f): Spline {
+class CatmullRomSpline(private val controlPoints: List<Vector3f>, val n: Int = 100, val alpha: Float = 0.5f,
```
skalarproduktraum, line 21: Have a close look here regarding what you want to have accessible and not. With the current declarations, controlPoints can't even be accessed from a derived class, while n and alpha are publicly accessible and can be changed!

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
```diff
@@ -14,24 +15,27 @@ import kotlin.math.pow
...
+                + (Pj.z()-Pi.z()).pow(2)).pow(exp) + ti)
     }
 
     /*
      * this function returns the spline points between two points. Please note you need four points
      * to have a smooth curve.
      * [n] is the number of points between the segments
      */
-    private fun CatmulRomSpline(p0: Vector3f, p1: Vector3f, p2: Vector3f, p3: Vector3f, n: Int = 100): List<Vector3f> {
+    private fun catmulRomSpline(p0: Vector3f, p1: Vector3f, p2: Vector3f, p3: Vector3f, n: Int = 100): List<Vector3f> {
```
skalarproduktraum, line 38: You robbed Ed Catmull of one `l`!

src/main/kotlin/graphics/scenery/DummySpline.kt
```diff
@@ -0,0 +1,27 @@
+package graphics.scenery
+
+import org.joml.*
```
skalarproduktraum, line 3: Maybe just import Vector3f here?

src/test/tests/graphics/scenery/tests/examples/basic/CurveCatmullRomExample.kt
```diff
@@ -1,17 +1,16 @@
...
-import graphics.scenery.CatmullRomSpline
-import org.joml.Vector3f
 
 import org.junit.Test
 
 /**
  * Just a quick example of a CatmullRomSpline with a triangle as a baseShape.
  *
- * @author Justin Bürger
+ * @author  Justin Buerger <burger@mpi-cbg.de>
```
skalarproduktraum, line 13: Just a note: Ümläüts are okay, as the source code is supposed to be UTF8.

src/test/tests/graphics/scenery/tests/unit/CurveTests.kt
```diff
@@ -1,21 +1,22 @@
...
-import graphics.scenery.CatmullRomSpline
-import graphics.scenery.Curve
-import graphics.scenery.UniformBSpline
+import graphics.scenery.*
 import graphics.scenery.utils.LazyLogger
-import org.joml.Vector3f
 import org.junit.Test
 import org.lwjgl.BufferUtils
+import kotlin.reflect.full.declaredMemberFunctions
+import kotlin.reflect.jvm.isAccessible
```
skalarproduktraum, line 10: These do not seem to be used anywhere, make sure you clean up imports.

skalarproduktraum (changes requested):
Looks good, but there are some minor things that need changes 👍

src/test/tests/graphics/scenery/tests/unit/CurveTests.kt
Domino2357: I believe they are necessary for the reflection - I access private functions in this test.

src/main/kotlin/graphics/scenery/CatmullRomSpline.kt
Domino2357: The poor guy!

src/main/kotlin/graphics/scenery/Axis.kt
Domino2357: Oh, ups, must have been my system 1 taking IntelliJs first suggestion.

skalarproduktraum:
@Domino2357 It looks like some of your tests are still sporadically failing, please make sure they work deterministically.

Domino2357:
Okay, appveyor found two NullPointerExceptions in the RibbonDiagramTests. I have no clue whatsoever where these come from because the tests don't fail on any other platform (IntelliJ, mvn test, gitlab, travis CI), plus, the visualizations for these entries are flawless.

Domino2357:
I would also be in favor of putting at least half of the 203 entries into a comment -as it is done currently- because this test would otherwise take too long.

skalarproduktraum (approved):

skalarproduktraum:
This looks great, thank you @Domino2357! 

Regarding the 203 proteins: I'd suggest picking maybe 20-50 of them at random, but outputting the seed value for the random number generator before choosing them. On the other hand this also destroys determinism a bit, so it might not be a good idea.
I will merge this anyway now, as I think this is not a blocking issue 👍

PR #339, opened 2020-10-25
GPU compatibility matrix

xulman:


xulman:
grr, why it does not take a default  name from the branch...

skalarproduktraum (approved):

PR #342, opened 2020-11-11
Assimp

elect86:
Long short story: prepare the terrain for Assimp:

* moved mesh loading functions to `object MeshImporter`
* refractored `Mesh` class itself and all its subclasses, including `MeshImporter`, under `/mesh`, less clutter

All the loading methods were previously belonging to the Mesh class, in order to create a mesh, setting some parameters (such as the name) and then continue with the loading. This was the case with OpenVR.

To solve this, I came up with appending to each `read*` method the resulting mesh

```
mesh: Mesh = Mesh()
```

In this way, in OpenVR nothings changes, you create your mesh in advance, set the name and pass it on.
For all the other cases, you simply

```
val mesh = MeshImporter.read*()
```

which will create the resulting instance automatically

This does not play perfectly whenever you want to create a class extending `Mesh`, such as `PointCloud`. You need, indeed, to instantiate the mesh manually by yourself in advance and then pass it down

```
    val pointCloud = PointCloud()
    MeshImporter.readFromOBJ(file, importMaterials = false, pointCloud)
```

We might change `read*` in order to make it play nice also in these situations, ie

```
    val pointCloud = MeshImporter.readFromOBJ<PointCloud>(file, importMaterials = false)
```

But this is fine tuning, I think

Ps: there are some gradle files, I messed up with the branches, but you can easily ignore them. I won't delete them just because they are going to be present sooner (hopefully) if my evil gradle migration plan goes through :)

Also, Assimp presents you with an [`AiScene`](http://assimp.sourceforge.net/lib_html/structai_scene.html) which may contain easily more than one mesh.

kephale smartly suggested to create a `Group` and add the meshes to this as children.

In order to accomodate that, we may still provide a convenient way for one-mesh situations:

* `MeshImporter` will attempt to load with Assimp and get only the very first mesh
* `GroupImporter` will do the proper thing

What do you think about it?


skalarproduktraum:
Thanks for the PR @elect86! 😎

A few things:

1. I'd actually like to keep the importing logic in Mesh, but probably move it to the companion object. Having a separate class just for importing doesn't seem to nice. 
2. In terms of importing children or not, I'd import them by default. Just importing the first Mesh is counterintuitive. There is also no need for additional Node types here, the parent can just be a Mesh, and all the sub-meshes from the import can be added to its children 👍
3. This PR includes all the Gradle stuff from #343, please don't mix PR contents! Please remove them from the PR, this one will (hopefully 😂) be merged faster than the Gradle one because the Gradle one definitely needs more testing in terms of everything playing nice with the other kids in the scijava sandbox 👍
4. I very much like the reorganisation and wanted to do this myself for some time already, but didn't get around it! Could you please separate the mesh stuff from the other primitives though? Such as in:
    * Sphere, Box, ... goes in `primitives/`
    * Mesh, etc goes in `mesh/`

elect86:
>     1. I'd actually like to keep the importing logic in Mesh, but probably move it to the companion object. Having a separate class just for importing doesn't seem to nice.

They aren't short and pollute the Mesh class, imho.
The critic on the separate class makes sense though.. what about then top level functions?
 
>     2. In terms of importing children or not, I'd import them by default. Just importing the first Mesh is counterintuitive. There is also no need for additional Node types here, the parent can just be a Mesh, and all the sub-meshes from the import can be added to its children 👍

Ok, so one root Mesh parent node and then all the meshes added as its children

>     3. This PR includes all the Gradle stuff from #343, please don't mix PR contents! Please remove them from the PR, this one will (hopefully 😂) be merged faster than the Gradle one because the Gradle one definitely needs more testing in terms of everything playing nice with the other kids in the scijava sandbox 👍

I'll do

>     4. I very much like the reorganisation and wanted to do this myself for some time already, but didn't get around it! Could you please separate the mesh stuff from the other primitives though? Such as in:
>        
>        * Sphere, Box, ... goes in `primitives/`
>        * Mesh, etc goes in `mesh/`

Sure, but what do you mean by `etc` other than `Mesh` for `mesh/`?

skalarproduktraum:
Closing, as outdated.

PR #343, opened 2020-11-11
gradle

elect86:
Maven to Gradle migration

Features:
- Gradle (mic drop.jpg)

This is almost complete (in b4 last famous words)

All the publishing stuff has been moved to a pre-compiled script plugin (`buildSrc/src/main/kotlin/publish.gradle.kts`) and applied in the main `build.gradle.kts` via 

    plugins {
       publish
       ...
    } 

Also, regarding the BOM, I initially started to do the right way, ie exploiting the official gradle support. However there are corner cases which messed it up, such as

        <dependency>
            <groupId>net.java.dev.jna</groupId>
            <artifactId>jna</artifactId>
        </dependency>

        <dependency>
            <groupId>net.java.dev.jna</groupId>
            <artifactId>jna-platform</artifactId>
            <version>${jna.version}</version>
        </dependency>

Then I went it hard and switched to the manual way, that is now I read the pom myself at begin, save it as a String in a variable and use pattern matching and regexes (I need to polish off some forgotten knowledge about) in order to extract what I want.

The cool thing is that now we have the full control in our hands. An example: gluegen use a slightly modified string to catch its version

        <dependency>
            <groupId>org.jogamp.gluegen</groupId>
            <artifactId>gluegen-rt</artifactId>
            <version>${gluegen-rt-main.version}</version>
        </dependency>

So I implemented in the following way: Gradle will first attempt to extract `<$artifactId.version>()</$artifactId.version>`, if it will fails, then it will fallback to a generic `<$artifactId*.version>()</$artifactId*.version>`, that is first `<gluegen-rt.version>()</gluegen-rt.version>` and then `<gluegen-rt*.version>()</gluegen-rt*.version>`, which will find a match in `gluegen-rt-main.version`

Also, it's recursive, so `${gluegen-rt-main.version}` will be properly handled

This flexibility offers us the option to push thing further toward simplifications. Let's take back the jna example, I can simply type

    sci("net.java.dev.jna:jna")
    sci("net.java.dev.jna:jna-platform:\$jna")

Because I know `$jna` will be resolved to `jna.version`

Also, I saw scenery was loading *all* the natives whenever they available. I also improved this and now only what needed for the current platform gets loaded

Legend:
- `sci` stays for "search the version in the sci pom"
- `testSci` is for testing if you can skiing


.mvn/.gradle-enterprise/gradle-enterprise-workspace-id
```diff
@@ -0,0 +1 @@
+juftkio6ovccraxoz7vxuvtuba
```
skalarproduktraum, line 1: What is this ID used for? Should it be private?

src/test/kotlin/graphics/scenery/tests/examples/basic/SponzaExample.kt
```diff
@@ -44,7 +44,7 @@ class SponzaExample : SceneryBase("SponzaExample", windowWidth = 1280, windowHei
```
skalarproduktraum, line 44: Those changes do not belong here.

buildSrc/src/main/kotlin/util.kt
```diff
@@ -0,0 +1,153 @@
...
+import org.gradle.api.artifacts.ExternalModuleDependency
+import org.gradle.api.artifacts.dsl.DependencyHandler
+import org.gradle.kotlin.dsl.DependencyHandlerScope
+import org.gradle.kotlin.dsl.accessors.runtime.addExternalModuleDependencyTo
+import org.gradle.nativeplatform.platform.internal.DefaultNativePlatform
+import java.net.URL
+import java.util.regex.Pattern
+
+// last commit of pom-scijava-29.2.1
+val pomSci = URL("https://raw.githubusercontent.com/scijava/pom-scijava/10b42e8e/pom.xml").readText()
```
skalarproduktraum, line 12: So this is the solution to using parent POMs in Gradle, nice! Can you make this more flexible so we can actually specify the POM version? Maybe you could get the XML actually from the Maven repo? E.g. from https://maven.scijava.org/content/groups/public/org/scijava/pom-scijava/29.2.1/pom-scijava-29.2.1.pom

skalarproduktraum (changes requested):
This looks pretty good! I have a few questions and changes, and I have not yet tested it myself. I will do that in the coming days.

skalarproduktraum:
Also another question: How does publishing with Gradle work, then? For Maven, we have everything including signing set up, so it's just necessary to run `mvn -Prelease deploy`. Can we get a similarly simple workflow with Gradle?

src/test/kotlin/graphics/scenery/tests/examples/basic/SponzaExample.kt
elect86: I honestly have no idea how they got there

.mvn/.gradle-enterprise/gradle-enterprise-workspace-id
elect86: According [here](https://docs.gradle.com/enterprise/maven-extension/):

> Since version 1.2, the Gradle Enterprise Maven extension captures an identifier used to uniquely represent a given workspace.
>    For versions 1.2.3+, this identifier is stored under .mvn/.gradle-enterprise/gradle-enterprise-workspace-id. The .mvn/.gradle-enterprise folder should NOT be committed under version control.

I'd remove it

elect86: Wait, let's make instead make `.gitignore` a badass one, I'll push a new one

buildSrc/src/main/kotlin/util.kt
elect86: ah, that's much better actually, thanks

.mvn/.gradle-enterprise/gradle-enterprise-workspace-id
elect86: pushed, no idea why is not updating here, however you can see it in the "commit" tab

src/test/kotlin/graphics/scenery/tests/examples/basic/SponzaExample.kt
elect86: addressed

buildSrc/src/main/kotlin/util.kt
kephale: Ok. I'll note that I sometimes use jitpack on pom-scijava versions so the commit hash fetching is kinda nice in that case. pom-scijava can have slow downs in the release process, which is why I started to do that. 

That said, it is also possible that maven.scijava.org's mirroring of jitpack would still work with the hash. Would need to check.

kephale:
Two things I am curious about:

1) Does this continue to work in jitpack as maven previously did? A relatively easy sanity check would be to use the hash of this PR in sciview and see what happens with different platforms. I could check this at some point, but it will be a bit because my queue is a bit overloaded.
2) I actually have a concern about removing *all* natives from scenery, but this might just be a clarification. Do you mean scenery is loading all natives at runtime? or is this about all natives getting bundled? I just want to make sure that when we depend on scenery it actually delivers *all* natives still, regardless of the platform that it was built on.

buildSrc/src/main/kotlin/util.kt
elect86: Addressed

elect86:
> 
> 
> Two things I am curious about:
> 
>     1. Does this continue to work in jitpack as maven previously did? A relatively easy sanity check would be to use the hash of this PR in sciview and see what happens with different platforms. I could check this at some point, but it will be a bit because my queue is a bit overloaded.
> 
>     2. I actually have a concern about removing _all_ natives from scenery, but this might just be a clarification. Do you mean scenery is loading all natives at runtime? or is this about all natives getting bundled? I just want to make sure that when we depend on scenery it actually delivers _all_ natives still, regardless of the platform that it was built on.

1. Absolutely, I have been using gradle and jitpack since an era
2. This is a nice question, for sure they get bundled. Regarding the loading, instead, I think it depends on the library itself.

Why would you want that (_all_ natives)?

kephale:
> > Two things I am curious about:
> > ```
> > 1. Does this continue to work in jitpack as maven previously did? A relatively easy sanity check would be to use the hash of this PR in sciview and see what happens with different platforms. I could check this at some point, but it will be a bit because my queue is a bit overloaded.
> > 
> > 2. I actually have a concern about removing _all_ natives from scenery, but this might just be a clarification. Do you mean scenery is loading all natives at runtime? or is this about all natives getting bundled? I just want to make sure that when we depend on scenery it actually delivers _all_ natives still, regardless of the platform that it was built on.
> > ```
> 
>     1. Absolutely, I have been using gradle and jitpack since an era
> 
>     2. This is a nice question, for sure they get bundled. Regarding the loading, instead, I think it depends on the library itself.
> 
> 
> Why would you want that (_all_ natives)?

Because for the Fiji plugin we ship 1 jar and provide a "just works" soluition. Within the imagej updater we don't have logic to check OS. In theory we could do this for the standalone version of sciview, but there are so many other things that we need to do in the meantime that I really don't want to break something that works just to reduce package size. In the timespan of the next couple of months, we're short on dev time, not disk space.

ctrueden:
> Within the imagej updater we don't have logic to check OS.

I didn't fully follow everything above, but: you can ship platform-specific natives wrapped as JARs in the `jars/<platform>` folder, and the Updater understands that, delivering only the matching platform's directory. And you can ship bare platform-specific native libraries in `lib/<platform>`, which the Updater understands. @kephale I know you already know this—I'm mostly saying it for @elect86's benefit—but also just curious what the dilemma is here, and if there's anything that could be done to help on the ImageJ side.

kephale:
@ctrueden I am sure everything is resolvable, but I have a concern that some of the changes will lead to OS-specific versions of scenery. I do not think this is an issue that is likely to require upstream changes. This stuff is not broken currently, and I am concerned they will become broken temporarily at a time when we cannot afford the time to fix things that wouldn't have needed to be fixed otherwise. I am happy to support the gradle changes and put time into testing things out later on.

elect86:
Then let's put this in standby for Kyle's concerns, since they make sense

I'll keep it in sync though

skalarproduktraum:
@elect86 @ctrueden @kephale I am slightly confused with the binaries, as Maven actually also only include the current platform and not all of them. In sciview, we solve this in the `populate_fiji.sh` script kinda manually: https://github.com/scenerygraphics/sciview/blob/master/populate_fiji.sh#L163

Bottom line, this is not a blocker for gradle.

skalarproduktraum:
@elect86 I also noticed that by default it's running both examples and unit tests. But default, it should only run the unit tests, and the examples should _never_ be run automatically. The only way they are run is by using the ExampleRunner as part of the `integration-tests` profile in a Maven build. Could you construct something similar for gradle?

kephale:
I'm trying to test out gradle https://github.com/kephale/sciview-gradle-example that depends on this PR but I have an issue (note that the sciview version was still built against a maven-based scenery):

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileKotlin'.
> Could not resolve all files for configuration ':compileClasspath'.
   > Could not download lwjgl-xxhash-3.2.3-natives-${scijava.platform.family.long}.jar (org.lwjgl:lwjgl-xxhash:3.2.3)
      > Could not get resource 'https://jcenter.bintray.com/org/lwjgl/lwjgl-xxhash/3.2.3/lwjgl-xxhash-3.2.3-natives-$%7Bscijava.platform.family.long%7D.jar'.
         > Could not GET 'https://jcenter.bintray.com/org/lwjgl/lwjgl-xxhash/3.2.3/lwjgl-xxhash-3.2.3-natives-$%7Bscijava.platform.family.long%7D.jar'. Received status code 400 from server: Bad Request
   > Could not download lwjgl-remotery-3.2.3-natives-${scijava.platform.family.long}.jar (org.lwjgl:lwjgl-remotery:3.2.3)
      > Could not get resource 'https://jcenter.bintray.com/org/lwjgl/lwjgl-remotery/3.2.3/lwjgl-remotery-3.2.3-natives-$%7Bscijava.platform.family.long%7D.jar'.
         > Could not GET 'https://jcenter.bintray.com/org/lwjgl/lwjgl-remotery/3.2.3/lwjgl-remotery-3.2.3-natives-$%7Bscijava.platform.family.long%7D.jar'. Received status code 400 from server: Bad Request

```

but it still tries to use the `scijava.platform.family.long` variable that is defined in `pom-scijava-base` even though scenery is the only file that explicitly depends on the lwjgl libs.

kephale:
Also note that this currently doesn't work with jitpack: https://jitpack.io/graphics/scenery/scenery/a196d73/build.log

elect86:
Making Travis uses Gradle:
- removed maven `Slf4jMavenTransferListener` 
- moved (open)jdk declaration into its own (global) section
- enabled artifacts caching
- commented debug osx code for showing available JDKs
- made `.gradlew` executable and added the scripts commands

PR #345, opened 2020-11-17
moved tests under standard location

elect86:
That is from `src/test/tests` to `src/test/kotlin`. 

Same as sources, which has already the default path.

Also fixed all the references in the docs


skalarproduktraum:
Thank you very much @elect86 👍

PR #347, opened 2020-11-24
Ribbon rainbow

Domino2357:
The ribbons should also be displayable with a rainbow colour map

### DIFF
--- a/src/main/kotlin/graphics/scenery/Curve.kt
+++ b/src/main/kotlin/graphics/scenery/Curve.kt
@@ -113,7 +113,8 @@ class Curve(spline: Spline, private val firstPerpendicularVector: Vector3f = Vec
             val remainder = partialCurveGeometry.size%sectionVertices
             val n = ((partialCurveGeometry.size)/sectionVertices).toInt()
             val add = if(n == 0) { 1 } else { remainder/n }
-            partialCurveGeometry.windowed(sectionVertices + add, sectionVertices + add, true) { section ->
+            partialCurveGeometry.windowed(sectionVertices + add, sectionVertices-1 + add, true)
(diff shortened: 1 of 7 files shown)


Domino2357:
Oh, maybe I should create a package for the protein stuff? Seems a bit inconvenient to have all these classes swimming in the main package, what do you think, @skalarproduktraum?

skalarproduktraum:
Yes, @Domino2357, putting them in a separate package seems reasonable. I think @elect86 has already looked into reorganising a bit, maybe we can continue this discussion here? I think stuff like Curve, CatmullRomSpline, etc could go into a package `geometry`, while the geometric primitives like Box, etc could go into `primitives`. Mesh things could go into `mesh`. Opinions? Maybe also from @kephale ?

elect86:
Mesh things are few (1, 2?), so I'd do as follows: a `graphics.scenery.mesh` root package with `geometry` and `primitives` underneath (which are logical subsets)

Domino2357:
@elect86 I like the idea! When @skalarproduktraum gives his approval, I'll do it

skalarproduktraum:
I think this is getting a bit too deep/segmented, how about putting the Mesh stuff in `geometry` as well, and have `primitives` on the same level?

skalarproduktraum:
I'd also be fine with having `geometry.primitives`

Domino2357:
Agreed! Maybe burying the primitives too deep is not an optimal idea.
I'd prefer to do another pull request for the packaging if that is alright

skalarproduktraum:
Yes, that sounds good. I'd merge that PR now then, and then you can do the magic sorting hat 👍

PR #349, opened 2020-11-27
Ruler stuff

Domino2357:
Here is the ruler stuff for scenery.


skalarproduktraum:
The `CreateCommandScreen` class that should probably be part of the PR is missing :(

PR #352, opened 2020-12-07
RibbonDiagramTests: removed problematic protein

Domino2357:
RibbonDiagramTests: removed problematic protein


skalarproduktraum:
Thanks @Domino2357 --- please add an issue though that the memory consumption or the Ribbon class definitely needs to be optimised! Also have a look with e.g. VisualVM what's eating all the resources there.

PR #360, opened 2020-12-15
Packaging

Domino2357:
Reordered packages in scenery. Hopefully, it makes the class structure a little clearer. Actually, skalarproduktraum do you think a gitbook article on the package structure in scenery would be useful?


skalarproduktraum:
@Domino2357 with respect to #371, can this one be closed?

PR #364, opened 2021-01-15
Bump dokka to 1.4.20 and make arrow be in code block

MarcinAman:
Small improvement for the arrow rendering, now it is a code block, so formatting will be preserved in dokka.

Also i've bumped dokka to 1.4.20 since it contains some bugfixes


skalarproduktraum:
Thanks, @MarcinAman 👍

PR #379, opened 2021-03-03
Colormap: add fromColor method

frauzufall:
This adds a simple method for creating a `Colormap` based on a single color. It linearly blends between black and the color, which is not completely natural, but useful. Happy for ideas for improvements!


skalarproduktraum:
Looks great, thanks @frauzufall 👍. Could you please add a test for that functionality though?

PR #382, opened 2021-03-16
Properly close / dispose scenery

frauzufall:
This PR should address issues preventing scenery / sciview to fully terminate.

It's a draft because I don't manage to properly test / pinpoint the issue. 
This doesn't terminate:
```
SciView sv = SciView.create();
sv.closeWindow();
sv.close();
sv.getScijavaContext().dispose();
sv.dispose();
```
It'd be great to first write a test in scenery showing how to close scenery properly. I'm starting with this PR because I did find some potentially blocking issues (see commits, happy for feedback!) and this draft can be used to discuss what's happening.


frauzufall:
@skalarproduktraum I found the issue, it's `VulkanRenderer` related. In this case `SceneryBase.close()` is actually never called when the windows closes. I can make it properly close now but that breaks the `OpenGLRenderer`, will now look into how to make both work.

skalarproduktraum:
This should have been addressed in the meantime, considering the PR is also quite outdated, I'll close it.

PR #386, opened 2021-03-24
VolumeManager/Volume: Fix issue with compute shaders

skalarproduktraum:
@aryaman-gupta had reported an issue where compute shaders would not be used if the VolumeManager is recreated. This PR fixes that, and introduces a new `customTextures` list in VolumeManager, where the developer can specify custom textures that are not under the VolumeManager's responsibility.


skalarproduktraum:
How is this looking @aryaman-gupta, can we merge this?

aryaman-gupta:
Yes @skalarproduktraum, looks good! It worked well with my insitu code 👍

PR #376, opened 2021-02-24
mouse dragging behavior

moreApi:
Adds a mouse dragging behavior which allows the user to drag objects with their mouse. An example is included. This behavior should complement the `ControllerDrag` behavior for handhelds.

PR #387, opened 2021-03-24
Ortho view

moreApi:
Orthogonal view and new slicing, cropping and combi modes for slicing planes

This branch depents on the #376 mouse drag PR


src/main/kotlin/graphics/scenery/volumes/Volume.kt
```diff
@@ -102,6 +102,23 @@ open class Volume(val dataSource: VolumeDataSource, val options: VolumeViewerOpt
     /** Plane equations for slicing planes mapped to origin */
     var slicingPlaneEquations = mapOf<SlicingPlane, Vector4f>()
 
+    /** Modes how assigned slicing planes interact with the volume */
+    var slicingMode = SlicingMode.NONE
+
+    enum class SlicingMode(val id: Int){
+        // Volume is rendered as it is
+        NONE(0),
```
skalarproduktraum, line 110: Please make the Enum members non-caps, so e.g. just `None`, `Cropping` etc 👍

skalarproduktraum (changes requested):
Minor case change, that's it 👍 Thanks!

src/main/kotlin/graphics/scenery/volumes/Volume.kt
moreApi: done

PR #391, opened 2021-03-29
New new packaging

Domino2357:
This is the new new packaging pull request. The old one has become unusable because I made a mistake in the rebasing process which was practically irreversible.

Should be ready to merge.


Domino2357:
Deprecated, see #394

PR #392, opened 2021-03-31
BioJava NullpointerException

Domino2357:
Added a catch for NullPointerException and updated BioJava. The latter seems to solve the issue, though we can't be certain since the error was occurring nondeterministically.
I was thinking about changing the error message to: "NullPointerException in the BioJava-Code occurred. Please update BioJava and see if this resolves the issue. If not, please contact burger@mpi-cbg.de" What do you think @skalarproduktraum ?

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -98,7 +98,7 @@ dependencies {
     implementation("com.github.LWJGLX:lwjgl3-awt:cfd741a6")
     sciJava("org.janelia.saalfeldlab:n5"["", "-imglib2"])
     listOf("core", "structure", "modfinder").forEach {
-        sciJava("org.biojava:biojava-$it:5.3.0") {
+        sciJava("org.biojava:biojava-$it:5.4.0") {
             exclude("org.slf4j", "slf4j-api")
             exclude("org.slf4j", "slf4j-simple")
             exclude("org.apache.logging.log4j", "log4j-slf4j-impl")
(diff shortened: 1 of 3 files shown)


skalarproduktraum:
Sorry, but I fundamentally disagree this is a good strategy :-/ You can't just say "oh yeah something went wrong, please update the biojava version" without having a clue _what actually went wrong_. Please try to run the tests until you've found what actually went wrong. Even if it is on the biojava side, we still need to know what's the issue -- even if it's just caused by a connection problem or something.

Domino2357:
Admittedly, that's a good point. I'll check.

Domino2357:
Okay, after digging deeper we know the following:
The NullPointerException occurs deterministically -- the indeterminism stems from us testing randomly 20 out of 80 entries. In the BioJava-Code the issue was a missing Null-safeguard against a residue-type being not defined (in MmtfStructureReader.java, line 217). This has been resolved in the latest stable version (5.4.0). Consequently, our dependency has been updated. The reason why this happened just recently and never before is a recent change in the rcsb database (https://www.rcsb.org/structure/6SLM, also screenshot).
image

**Bottom-Line** the issue was an update to the database leading to a NullPointerException in BioJava and is resolved in the (new) version 5.4.0.

Domino2357:
@skalarproduktraum I think this is good to go.

skalarproduktraum:
Thanks @Domino2357 for figuring this out!

Issue #390, opened 2021-03-29
NullPointerException in RibbonDiagram


skalarproduktraum:
This seems to happen not really deterministically, could you please have a look, @Domino2357? 
See e.g. the test log at https://github.com/scenerygraphics/scenery/runs/2193828815#step:6:127

skalarproduktraum:
Another one, same line, log at https://ci.appveyor.com/project/skalarproduktraum/scenery/builds/38455773

Domino2357:
See #392

PR #394, opened 2021-04-01
New packagin

Domino2357:
This should do the job and solve the issue with @frauzufall branch. I merged the master into this one, thus, the merging into master should go smooth. Also, I did not change things besides moving things into different packages and deleting one minor test class I wrote.


Domino2357:
@frauzufall does the rebase on this work?
Otherwise, this should be ready to merge -- the build in Linux fails because of #390 which is taken care of by #392.

frauzufall:
@Domino2357 something's still off and I can't rebase. If you look into the commit list here, 9 days ago all commits were added a second time somehow. The last commit in the history on my branch from `newPackagin` is `753651d8` - which is exactly the commit prior to the point where all commits appear a second time. Can we reset to this commit and merge master again? Happy to do that with you / talk it through!

Domino2357:
@frauzufall I reverted and fixed some import-issues. Would you mind checking if you can rebase? If so, this should be ready to go.

Domino2357:
Heyho @frauzufall, the rebase is not possible automatically, however, from the first glance it seems we are only dealing with some renaming issues. If that's correct, this should be easy to resolve. Happy to talk to you if that seems helpful to you.

skalarproduktraum:
Closed because superceeded by #411.

Issue #395, opened 2021-04-01
Merge post render lambdas


kephale:
We need the post render lambdas from @aryaman-gupta's variant. We need this to ensure synchronization on a per-rendered-frame basis for things like 3Dscript integration in sciview.

PR #396, opened 2021-04-02
Post render functionalities for synchronized animations and texture requests

aryaman-gupta:
Adds capability to insert a user-defined lambda at the end of a render loop, which can be used to create consistent animations that are synchronized with the rendering. `PostRenderLambdaExample` shows how the functionality can be used. Also adds capability for persistent texture requests which are served once per render frame. These may particularly be useful, e.g, when compute shaders are used and output textures need to be accessed every time they are updated. The `PersistentTextureRequestsExample` shows how the functionality may be used.

Once @kephale has a look at whether this addresses #395, and @skalarproduktraum has a look at whether the code looks alright, I will add the same functionality to the `OpenGLRenderer` as well, thereby finalizing the PR.


src/main/kotlin/graphics/scenery/backends/Renderer.kt
```diff
@@ -175,6 +176,7 @@ abstract class Renderer : Hubable {
     }
 
     @Volatile protected var textureRequests = ConcurrentLinkedQueue<Pair<Texture, Channel<Texture>>>()
+    var persistentTextureRequests = ArrayList<Pair<Texture, AtomicInteger>>()
```
skalarproduktraum, line 179: Should be a `val` instead, reassignment might make more trouble than is solves

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
```diff
@@ -328,6 +329,7 @@ open class VulkanRenderer(hub: Hub,
     private var recordMovie: Boolean = false
     private var recordMovieOverwrite: Boolean = false
     override var pushMode: Boolean = false
+    val postRenderLambdas = ArrayList<()->Unit>()
```
skalarproduktraum, line 332: I know it was my naming originally, but postRenderLambdas is a bit cumbersome. Maybe we can rename them to something nicer?

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
```diff
@@ -380,7 +382,7 @@ open class VulkanRenderer(hub: Hub,
     var fps = 0
         protected set
     protected var frames = 0
-    protected var totalFrames = 0L
+    var totalFrames = 0L
```
skalarproduktraum, line 385: Should be `private set`.

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
```diff
@@ -1470,6 +1487,10 @@ open class VulkanRenderer(hub: Hub,
             }
         }
 
+        postRenderLambdas.forEach{
```
skalarproduktraum, line 1490: Missing space

src/test/kotlin/graphics/scenery/tests/examples/advanced/PersistentTextureRequestsExample.kt
```diff
@@ -0,0 +1,126 @@
...
+
+            // this is the loop where you may do your tasks that are asynchronous to rendering
+            while (renderer?.shouldClose == false) {
+                while (counter.get() == prevCounter) {
+                    Thread.sleep(5)
+                }
+
+                //updated texture has been returned
+                val buffer = opTexture.contents
+                //the buffer can now, e.g., be transmitted, as is required for parallel rendering
```
skalarproduktraum, line 103: Maybe use the buffer as a texture for another object here, for demo purposes?

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
skalarproduktraum: Also, it should probably be moved to `Renderer`.

skalarproduktraum (changes requested):
Only a few minor changes, otherwise looking good :+1

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
aryaman-gupta: I actually quite like the name, from a user's point of view. Makes it quite clear what the functionality is. Do you have another suggestion?

src/test/kotlin/graphics/scenery/tests/examples/advanced/PersistentTextureRequestsExample.kt
aryaman-gupta: A better example would be to write the texture to disk. Will that create any problems in the CI?
Using it as a texture in another object is not such a great example because you can do that without fetching the texture from the GPU

src/main/kotlin/graphics/scenery/backends/opengl/OpenGLRenderer.kt
```diff
@@ -2213,11 +2213,16 @@ open class OpenGLRenderer(hub: Hub,
...
+        postRenderLambdas.forEach {
+            it.invoke()
+        }
+
         stats?.add("Renderer.${flow.last()}.renderTiming", System.nanoTime() - startPass)
 
         updateLatch?.countDown()
         firstImageReady = true
         frames++
+        totalFrames++
```
aryaman-gupta, line 2225: @skalarproduktraum I have generalized the totalFrames parameter to the OpenGLRenderer, as was your suggestion. Is the OpenGLRenderer recreated at any point, which would then require totalFrames to be set to 0 again? I could not find any such recreation

src/test/kotlin/graphics/scenery/tests/examples/compute/PersistentTextureRequestsExample.kt
```diff
@@ -0,0 +1,143 @@
...
+
+            scene.addChild(this)
+        }
+
+        thread {
+            val opTexture = volumeManager.material.textures["OutputRender"]!!
+
+            var prevCounter = 0
+
+            (renderer as? VulkanRenderer)?.persistentTextureRequests?.add(opTexture to counter)
```
aryaman-gupta, line 94: @skalarproduktraum as we have discussed, the `PersistentTextureRequests` are implemented only in VulkanRenderer, as the OpenGLRenderer does not support compute shaders anyway. Should this example therefore be put in the `blocklist` of the ExampleRunner?

src/test/kotlin/graphics/scenery/tests/examples/compute/PersistentTextureRequestsExample.kt
```diff
@@ -0,0 +1,143 @@
...
+
+        thread {
+            while (renderer?.firstImageReady == false) {
+                Thread.sleep(5)
+            }
+
+            Thread.sleep(1000) //give some time for the rendering to take place
+
+            renderer?.close()
+            Thread.sleep(200) //give some time for the renderer to close
```
aryaman-gupta, line 121: @skalarproduktraum is there a way to query whether the renderer has terminated? Here, if the renderer does not close within 200 ms, the value of totalFrames recorded is incorrect. This leads to the test sometimes failing on CI, depending on whether the renderer closed in 200 ms or not

src/main/kotlin/graphics/scenery/backends/opengl/OpenGLRenderer.kt
skalarproduktraum: No, no such things happens for OpenGL 👍

src/test/kotlin/graphics/scenery/tests/examples/compute/PersistentTextureRequestsExample.kt
skalarproduktraum: Not necessary to add it to the blocklist, because the CI only runs the VulkanRenderer anyway.

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
aryaman-gupta: @skalarproduktraum please let me know if it is good as is so we can merge this

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
aryaman-gupta: Changed to `protected set` and moved to `Renderer`. `private set` does not compile

skalarproduktraum:
Superseded by #622.

PR #406, opened 2021-04-13
Refactor Node properties into attribute classes

frauzufall:
This branch makes `Node` an instance and sorts some it's properties into attribute classes (`Renderable`, `Material`, `Spatial`, `Geometry`). This design is extendable - one can also attach custom attribute classes to `Node` dynamically. See `AttributesExample` for details.


frauzufall:
@skalarproduktraum @elect86 @moreApi I updated this branch - these are the changes to what we looked at earlier today:
- properties are now called `attribute` as @kephale suggested
- I removed the unused attributes from `Spatial` and moved some of the attributes from `Renderable` to `Node` as discussed
- the whole logic of lazy attaching attributes to a node didn't actually work :see_no_evil: :see_no_evil: :see_no_evil: Now the attributes are added explicitly in `init` of a node, which removes some of the ugly corners. `Node` now also e.g. only implements `geometryOrNull()` and `ifGeometry{}`, and only the nodes extending  `HasGeometry` implement `geometry{}`. That's cleaner. Have a look at the updated `AttributesExample` class.
- I was not able to simplify `AttributesMap` - I don't know how to remove the class type parameter while still being able to use the class type as a key for the map. Doesn't work with the interfaces, happy for more ideas - I don't think it's relatable to `Settings` because there a string is used as a key in the map. 
- the `SelectionTest` fails - I don't know why yet. The `MouseInputExample` seems to work fine.

While

elect86:
Does the `SelectionTest` fail on your machine?

moreApi:
There is a strange error in `CroppingExample` when atleast two extra volumes are present. I am not sure how to tacke it. The logs spams somthing like:

```
[main] WARN VulkanObjectState - graphics.scenery.backends.vulkan.VulkanObjectState@1e90939f: Could not find descriptor set for geometry and texture set transferFunction_x_191_x_
[main] WARN VulkanObjectState - DS are: geometry: volume_x_158_x_, volumeCache, volume_x_174_x_, volume_x_190_x_, ObjectTextures, lights: volumeCache, volume_x_158_x_, volume_x_174_x_, volume_x_190_x_
[main] ERROR VulkanScenePass - Internal consistency error for node Volume (RAI source): Descriptor set transferFunction_x_191_x_ not found in renderpass ForwardShading, skipping node for rendering.
```

skalarproduktraum:
Thank you for taking on the epic quest of introducing the attribute system, @frauzufall! I'll merge this now and am looking forward to the cool things we can do with it 🥳

PR #414, opened 2021-04-27
Space filling diagram

Domino2357:
Provides a space filling visualization of proteins as well as an optimized ball and stick representation.


skalarproduktraum:
Is this ready and needs review or is there still work to be done?

Domino2357:
Needs a review, please :) @skalarproduktraum

src/main/kotlin/graphics/scenery/ChemicalElement.kt
```diff
@@ -0,0 +1,28 @@
...
+                           val atomicRadius: Float?,
+                           val ionizationEnergy: Float?,
+                           val electronAffinity: Float?,
+                           val oxidationStates: String,
+                           val standardState: String,
+                           val meltingPoint: Float?,
+                           val boilingPoint: Float?,
+                           val density: Float?,
+                           val groupBlock: String,
+                           val yearDiscovered: String)
```
skalarproduktraum, line 27: ಠ_ಠ

src/main/kotlin/graphics/scenery/PeriodicTable.kt
```diff
@@ -0,0 +1,111 @@
...
+    data class ChemTable(@JsonProperty("Columns") val columns: ChemColumn,@JsonProperty("Row") val row: ArrayList<ChemCell>)
+    data class PeriodicTableau(@JsonProperty("Table") val table: ChemTable)
+
+    val elementList = ArrayList<ChemicalElement>()
+    private val logger by LazyLogger()
+
+    init {
+        //parsing the json file
+        val mapper = jacksonObjectMapper()
+        val file = FileReader("src/main/resources/graphics/scenery/proteins/PubChemElements_all.json").readText()
```
skalarproduktraum, line 40: This won't work if things are packaged. Use `this::class.java.getResource(filename)` instead!

src/main/kotlin/graphics/scenery/PeriodicTable.kt
```diff
@@ -0,0 +1,111 @@
...
+            pureStringElement.cell[4].chunked(2).forEachIndexed { index, co ->
+                val colorValue = co.toInt(16).toFloat()/255f
+                when (index) {
+                    0 -> color.x = colorValue
+                    1 -> color.y = colorValue
+                    2 -> color.z = colorValue
+                }
+            }
+            val element = ChemicalElement(
+            pureStringElement.cell[0].toInt(),
```
skalarproduktraum, line 63: Could you use named arguments here? This is super-hard to read otherwise.

src/main/kotlin/graphics/scenery/StickAndBallProteinModel.kt
```diff
@@ -0,0 +1,134 @@
...
+import org.biojava.nbio.structure.*
+
+/**
+ * This is a representation of a protein structure. Displayed are the covalent bonds of the protein with atoms being
+ * represented with Icospheres and cylinders for the connections between them.
+ * @param [protein] Protein which is to be visualized.
+ *
+ * @author  Justin Buerger <burger@mpi-cbg.de>
+ */
+class StickAndBallProteinModell(protein: Protein, spaceFilling: Boolean = false,
```
skalarproduktraum, line 14: Typo

src/main/kotlin/graphics/scenery/StickAndBallProteinModel.kt
```diff
@@ -0,0 +1,134 @@
...
+                                displayExternalMolecules: Boolean = false): Mesh("PrimaryStructure") {
+    val structure = protein.structure
+    private val periodicTable = PeriodicTable()
```
skalarproduktraum, line 17: The periodic table will probably be used only as one instance for all, right? If so, move to `companion object`.

src/main/kotlin/graphics/scenery/PeriodicTable.kt
skalarproduktraum: Also, does that JSON file come with any license attached?

skalarproduktraum (changes requested):

src/main/kotlin/graphics/scenery/PeriodicTable.kt
Domino2357: Done.

There is no license- they kindly ask for a citation, though (https://pubchemdocs.ncbi.nlm.nih.gov/citation-guidelines). I will include this in the doc.

src/main/kotlin/graphics/scenery/PeriodicTable.kt
```diff
@@ -37,7 +40,7 @@ open class PeriodicTable {
     init {
         //parsing the json file
         val mapper = jacksonObjectMapper()
-        val file = FileReader("src/main/resources/graphics/scenery/proteins/PubChemElements_all.json").readText()
+        val file = this::class.java.getResource("src/main/resources/graphics/scenery/proteins/PubChemElements_all.json").readText()
```
skalarproduktraum, line 43: That again is not going to work when deployed as a JAR: The paths given to `getResource` are always relative to the class it's being called on. So if your `PubChemElements_all.json` is in the same package as the `PeriodicTable` class, all you need there is the file name 👍

src/main/kotlin/graphics/scenery/StickAndBallProteinModel.kt
```diff
@@ -11,10 +11,10 @@ import org.biojava.nbio.structure.*
  *
  * @author  Justin Buerger <burger@mpi-cbg.de>
  */
-class StickAndBallProteinModell(protein: Protein, spaceFilling: Boolean = false,
-                                displayExternalMolecules: Boolean = false): Mesh("PrimaryStructure") {
+class StickAndBallProteinModel(protein: Protein, spaceFilling: Boolean = false,
+                               displayExternalMolecules: Boolean = false): Mesh("PrimaryStructure") {
     val structure = protein.structure
-    private val periodicTable = PeriodicTable()
+    companion object PerTab private val periodicTable = PeriodicTable()
```
skalarproduktraum, line 17: I didn't know this syntax for companion object even exists 🤣 Could you, for consistency with other code, please enclose the objects in curly braces?

skalarproduktraum (changes requested):
Two minor things, otherwise looking very good. Please fix these and then we're ready to go 👍

src/main/kotlin/graphics/scenery/PeriodicTable.kt
Domino2357: Strangely enough it still doesn't work when I put both in the same package. getResource() still throws a NullPointerException

Domino2357:
Closed with regards to #428

PR #417, opened 2021-05-07
OrthoView.kt: add JvmName for OrthoView

kephale:
This adds a JvmName to OrthoView so it can be referenced from Java and Python (via scyjava) easily.


skalarproduktraum:
I'll merge this for the time being, we should have a discussion with @moreApi nevertheless how the OrthoView API could maybe be improved 👍

PR #418, opened 2021-05-10
Fix BoundingBox Bug for Proteins

### LINKED ISSUE #413: Bounding boxes are not generated correctly for RibbonDiagrams
### Steps to reproduce
Call `getMaximumBoundingBox()` on a `RibbonDiagram` instance, e.g. created via `val diagram = RibbonDiagram(Protein.fromID("2zzw"))`.

### Expected result
The function should return a non-empty bounding box.

### Actual result
An empty bounding box, with min and max containing the null vector, is returned.

Domino2357:
This fixes #413. The problem was that the positions were not well defined for subCurves in the Curve class.

### DIFF
--- a/src/test/kotlin/graphics/scenery/tests/unit/RibbonDiagramTests.kt
+++ b/src/test/kotlin/graphics/scenery/tests/unit/RibbonDiagramTests.kt
@@ -138,29 +138,30 @@ class RibbonDiagramTests {
         val protein = Protein.fromID("2zzw")
         val ribbon = RibbonDiagram(protein)
         val bb = ribbon.getMaximumBoundingBox()
+        assertEquals(bb.n, ribbon)
         assertNotEquals(bb.min, Vector3f(0f, 0f, 0f))
         assertNotEquals(bb.max, Vector3f(0f, 0f, 0f))
-        assertEquals(bb.n, ribbon)
     }
 
     /**
-     * Verifies that the correct BoundingBox is created.
+     * Verifies that a BoundingBox for a ribbon can be created.
      */
-
     @Test
-    fun testMaxBoundBox() {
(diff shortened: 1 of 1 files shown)


Domino2357:
@skalarproduktraum this should be fine.

Domino2357:
The failing tests do not seem to occur because of the one line in Curve changed.

skalarproduktraum:
@Domino2357 The CameraTests issue is again something OpenCL-related. However, the rendering errors for proteins in the Argos checks are not: https://www.argos-ci.com/scenerygraphics/scenery/builds/290 Could you please have a look and see what's the problem there? Thanks!

skalarproduktraum:
Any news here, @Domino2357?

Domino2357:
Thanks for checking in @skalarproduktraum ! I don't know how to assign the right positions for the subcurves. All vertices are stored correctly, but I never specified a position. Naively, its just the centroid of all vertices but this breaks the rendering. Also, the first splinePoint, i.e., the beginning of each subcurve deosn't work either. Is there a way to specify the position when all vertex positions are already in place?

skalarproduktraum:
Not sure I understand the question, @Domino2357 😵 What do you mean by "Is there a way to specify the position when all vertex positions are already in place?" -- do you want to modify the vertex buffer without rewriting everything?

Domino2357:
@skalarproduktraum sorry that was put clumsy. My problem is once I defined a position for each subcurve, the rendering breaks. I tried to use the centroid of the vertices as the position as well as the starting point of the sub-spline. My question is: how can one define a position without changing the vertices positions?

skalarproduktraum:
@Domino2357 This PR would need a merge of master 👍

PR #422, opened 2021-05-17
VulkanRenderer: Fix issue where screenshots would not be overwritten

### LINKED ISSUE #421: screenshot does not overwrite previous file when using VulkanRenderer
When overwrite = true with VulkanRenderer the destination file is not overwritten. When OpenGLRenderer is used the file is correctly overwritten.

skalarproduktraum:
This PR fixes #421, where screenshots were not overwritten as requested when using the Vulkan renderer.


skalarproduktraum:
@kephale Any feedback on this?

kephale:
Looks good to me!

PR #426, opened 2021-06-02
Orthoview improvements for volume visualisation

moreApi:
Depends on #406 and #411.

- Add node filter to raycast collision detection as second mode instead of ignoring classes.
- OrthoView Input behavior uses this new filtering.
- OrthoView input behavior now adds itself automatically.


moreApi:
@skalarproduktraum says, it belongs in scenery

moreApi:
@skalarproduktraum it is green now! It would be good if the review would happen before the logs get deleted because of old age and the PR goes red again because of missing logs.

skalarproduktraum (approved):

PR #431, opened 2021-06-18
imgui support

skalarproduktraum:
This is excellent work, @elect86! I hope you don't mind me opening the PR here so we can discuss.

There's still a few issues that need to be ironed out. The generic ones I'll point out here, the specific ones in the code.

- [ ] Scaling on non-hidpi screens is incorrect, imgui widgets are twice the size they should be
- [ ] Mouse events bubble through, so the camera is also moved when an imgui widget is moved
- [ ] Resizing imgui windows doesn't work
- [ ] non-GLFW-backed windows need to work
 
Support for non-GLFW is unclear. We probably do need some additional glue for this to work e.g. on Swing, right?


build.gradle.kts
```diff
@@ -97,6 +97,13 @@ dependencies {
     testImplementation(imagej.core)
     testImplementation(imagej.ij)
     testImplementation(imgLib2.ij)
+
```
skalarproduktraum, line 100: Why do we need all of those libs? Aren't they at least partially transitive deps of your imgui port?

src/main/kotlin/graphics/scenery/ui/menu.kt
```diff
@@ -0,0 +1,40 @@
+package graphics.scenery.ui
+
+import glm_.vec2.Vec2
+import graphics.scenery.Node
+import graphics.scenery.Scene
+import imgui.dsl
+import imgui.ImGui
+import java.text.SimpleDateFormat
+
+object menu {
```
skalarproduktraum, line 10: Wrong case

skalarproduktraum: Also the name could be a bit more specific :-) Maybe that code should actually be moved into the example?

skalarproduktraum (commented):
There are also a bunch of code style changes that move away from the one that is generally used in scenery. Please remove them, especially in `Node`. I know we need a style file very soon.

skalarproduktraum:
@elect86 How shall we continue with this PR? Have you worked on this lately?

elect86:
I've not been working on, since Where2Test almost fully took all my capacity, I'll refresh imgui and fix the  conflicts

elect86:
So, to give this super priority, I create a uber-jar, added it locally and it worked out-of-the-box

The `ui-example` runs flawless (although without sharpness)

PR #436, opened 2021-07-29
Added parser for .cube files (DFT output, atomic positions + electronic density)

RandomDefaultUser:
This PR adds functionality (+examples) to parse .cube files. These are commonly used output files from density functional theory, holding atomic positions and electronic density. The implemented parser class can be extended in the future to allow input from a wide array of DFT output data. The parsed data can be used to visualize a DFT-MD.

There are still some edges before this can be merged: 

1. I am not sure if I put the parser and examples in the correct directory. 
2. When visualizing small unit cells, one of the edges seems to be wrongly zero (or close to it). I do not see this when visualizing larger cells and I've also checked the buffer for values that never get assigned. I will investigate further, but if anyone has a suggestion, I'd be thankful. (see picture)
Screenshot from 2021-07-29 19-37-46

3. I'd like to add my demo data to the demo data .zip file. Locally, I've done so to test everything.

### DIFF
--- a/src/main/kotlin/graphics/scenery/DFTParser.kt
+++ b/src/main/kotlin/graphics/scenery/DFTParser.kt
@@ -1,45 +1,46 @@
 package graphics.scenery
+import graphics.scenery.repl.REPL
 import graphics.scenery.utils.RingBuffer
 import org.joml.Vector3f
 import org.lwjgl.system.MemoryUtil
 import java.io.File
 import java.nio.ByteBuffer
 
-class DFTParser (val fileType: String = "cube", val normalizeDensityTo: Float = -1.0f){
-    /*
-     Properties.
-     */
-    var numberOfAtoms:Int = 0
-    var gridSpacings = FloatArray(3) { 0.0f }
-    var gridDimensions = IntArray(3) { 0 }
-    var atomicPositions = Array(0){ Vector3f()}
-    var electronicDensity = Array(0, { Array(0, { FloatArray(0) } ) } )
-    lateinit var electronicDensityUInt : ByteBuffer
+/**
+Parses a density functional theory (common simulation method in solid state physics and theoretical chemistry)
+calculation. Can be used to visualize single DFT calculations or DFT-MD (=DFT molecular dynamics simulations).
 
+ * @property[normalizeDensityTo] Defines to which value the density is scaled. This is useful when visualizing more then one DFT calculation at the
+    same time, in order to keep the density visualization consistent. Negative values mean the density is scaled
+    per snapshot. Default is -1.0f.
+ */
+class DFTParser (val normalizeDensityTo: Float = -1.0f){
+    /**  Number of Atoms.*/
+    var numberOfAtoms:Int = 0
+    /** Distance between two grid points in either direction in Bohr.*/
+    var gridSpacings = FloatArray(3) { 0.0f }
+    /** Number of gridpoints in 3D grid.*/
+    var gridDimensions = IntArray(3) { 0 }
+    /** Positions of the atoms in Bohr.*/
+    var atomicPositions = Array(0){ Vector3f()}
+    private var electronicDensity = Array(0, {Array(0, { FloatArray(0) } ) } )
+    /** Electronic density as scaled bytes, in scaled(1/Bohr).*/
+    lateinit var electronicDensityUInt : ByteBuffer
+    /** Origin of the unit cell, usually 0,0,0. */
     var unitCellOrigin = FloatArray(3) { 0.0f }
+    /** Dimensions of the unit cell, in Bohr.*/
     var unitCellDimensions = FloatArray(3) { 0.0f }
 
-    /*
-     Functions.
-     */
-
-    // Parse a file containing visualize-able information about a DFT calculation.
-    fun parseFile(filename: String) {
-        if (this.fileType == "cube") {
-            this.parseCube(filename)
-        }
-    }
-
     // Parse information as .cube file.
-    private fun parseCube(filename: String){
+    fun parseCube(filename: String){
         // Read entire file content
         val cubeFile =  File(filename).bufferedReader().readLines()
         var counter = 0
         var xcounter = 0
         var ycounter = 0
         var zcounter = 0
-        var minDensity:Float = 1000.0f
-        var maxDensity:Float = 0.0f
+        var minDensity = 1000.0f
+        var maxDensity = 0.0f
 
         // Iterate through file. We know what is were with a cube file.
         // 0-1: Comments.
@@ -112,7 +113,7 @@ class DFTParser (val fileType: String = "cube", val normalizeDensityTo: Float = 
         // Converting to byte buffer.
         counter = 0
         electronicDensityUInt =  MemoryUtil.memAlloc((gridDimensions[0]*
-            gridDimensions[1]*gridDimensions[2]*1).toInt())
+            gridDimensions[1]*gridDimensions[2]*1))
         if (normalizeDensityTo > 0.0f){
             maxDensity = normalizeDensityTo
         }
--- a/src/main/kotlin/graphics/scenery/DFTParser.kt
+++ b/src/main/kotlin/graphics/scenery/DFTParser.kt
@@ -5,7 +5,7 @@ import org.lwjgl.system.MemoryUtil
 import java.io.File
 import java.nio.ByteBuffer
 
-class DFTParser (val fileType: String = "cube"){
+class DFTParser (val fileType: String = "cube", val normalizeDensityTo: Float = -1.0f){
     /*
      Properties.
      */
@@ -113,7 +113,9 @@ class DFTParser (val fileType: String = "cube"){
         counter = 0
         electronicDensityUInt =  MemoryUtil.memAlloc((gridDimensions[0]*
             gridDimensions[1]*gridDimensions[2]*1).toInt())
-
+        if (normalizeDensityTo > 0.0f){
+            maxDensity = normalizeDensityTo
+        }
         for (z in 0 until gridDimensions[2]){
             for (y in 0 until gridDimensions[1]){
                 for (x in 0 until gridDimensions[0]){
--- a/src/main/kotlin/graphics/scenery/DFTParser.kt
+++ b/src/main/kotlin/graphics/scenery/DFTParser.kt
@@ -1,6 +1,9 @@
 package graphics.scenery
+import graphics.scenery.utils.RingBuffer
 import org.joml.Vector3f
+import org.lwjgl.system.MemoryUtil
 import java.io.File
+import java.nio.ByteBuffer
 
 class DFTParser (val fileType: String = "cube"){
     /*
@@ -11,6 +14,8 @@ class DFTParser (val fileType: String = "cube"){
     var gridDimensions = IntArray(3) { 0 }
     var atomicPositions = Array(0){ Vector3f()}
     var electronicDensity = Array(0, { Array(0, { FloatArray(0) } ) } )
+    var electronicDensityUInt = RingBuffer<ByteBuffer>(1) { MemoryUtil.memAlloc((0).toInt()) }
+
     var unitCellOrigin = FloatArray(3) { 0.0f }
     var unitCellDimensions = FloatArray(3) { 0.0f }
 
@@ -33,6 +38,8 @@ class DFTParser (val fileType: String = "cube"){
         var xcounter = 0
         var ycounter = 0
         var zcounter = 0
+        var minDensity:Float = 1000.0f
+        var maxDensity:Float = 0.0f
 
         // Iterate through file. We know what is were with a cube file.
         // 0-1: Comments.
@@ -76,16 +83,22 @@ class DFTParser (val fileType: String = "cube"){
                     }
 
                     // Parsing volumetric data.
-                    if (counter >= 6+numberOfAtoms){
-                        for (value in (line.trim().split("\\s+".toRegex())))
-                        {
+                    if (counter >= 6+numberOfAtoms) {
+                        for (value in (line.trim().split("\\s+".toRegex()))) {
                             // Cube files should be in Fortran (z-fastest ordering).
                             electronicDensity[xcounter][ycounter][zcounter] = (value).toFloat()
+                            if (electronicDensity[xcounter][ycounter][zcounter] > maxDensity) {
+                                maxDensity = electronicDensity[xcounter][ycounter][zcounter]
+                            }
+                            if (electronicDensity[xcounter][ycounter][zcounter] < minDensity) {
+                                minDensity = electronicDensity[xcounter][ycounter][zcounter]
+                            }
+
                             zcounter++
-                            if (zcounter == gridDimensions[2]){
+                            if (zcounter == gridDimensions[2]) {
                                 zcounter = 0
                                 ycounter++
-                                if (ycounter == gridDimensions[1]){
+                                if (ycounter == gridDimensions[1]) {
                                     ycounter = 0
                                     xcounter++
                                 }
@@ -96,5 +109,22 @@ class DFTParser (val fileType: String = "cube"){
             }
             counter++
         }
+        // Converting to byte buffer.
+        // This is all very hacky and just for testing purposes.
+        counter = 0
+
+        electronicDensityUInt = RingBuffer<ByteBuffer>(1) { MemoryUtil.memAlloc((gridDimensions[0]*
+            gridDimensions[1]*gridDimensions[2]).toInt()) }
+
+        for (x in 0 until gridDimensions[0]){
+            for (y in 0 until gridDimensions[1]-1){
+                for (z in 0 until gridDimensions[2]-1){
+                    val value:Int = (((electronicDensity[x][y][z]-minDensity) / (maxDensity-minDensity))*255.0f).toInt()
+                    electronicDensityUInt.get().put(value.toByte())
+                    counter++
+                }
+            }
+        }
+
     }
 }
--- a/src/main/kotlin/graphics/scenery/DFTParser.kt
+++ b/src/main/kotlin/graphics/scenery/DFTParser.kt
@@ -1,0 +1,100 @@
+package graphics.scenery
+import org.joml.Vector3f
+import java.io.File
+
+class DFTParser (val fileType: String = "cube"){
+    /*
+     Properties.
+     */
+    var numberOfAtoms:Int = 0
+    var gridSpacings = FloatArray(3) { 0.0f }
+    var gridDimensions = IntArray(3) { 0 }
+    var atomicPositions = Array(0){ Vector3f()}
+    var electronicDensity = Array(0, { Array(0, { FloatArray(0) } ) } )
+    var unitCellOrigin = FloatArray(3) { 0.0f }
+    var unitCellDimensions = FloatArray(3) { 0.0f }
+
+    /*
+     Functions.
+     */
+
+    // Parse a file containing visualize-able information about a DFT calculation.
+    fun parseFile(filename: String) {
+        if (this.fileType == "cube") {
+            this.parseCube(filename)
+        }
+    }
+
+    // Parse information as .cube file.
+    private fun parseCube(filename: String){
+        // Read entire file content
+        val cubeFile =  File(filename).bufferedReader().readLines()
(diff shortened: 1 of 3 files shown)


src/main/kotlin/graphics/scenery/DFTParser.kt
```diff
@@ -0,0 +1,132 @@
...
+            counter++
+        }
+        // Converting to byte buffer.
+        counter = 0
+        electronicDensityUInt =  MemoryUtil.memAlloc((gridDimensions[0]*
+            gridDimensions[1]*gridDimensions[2]*1))
+        if (normalizeDensityTo > 0.0f){
+            maxDensity = normalizeDensityTo
+        }
+        for (z in 0 until gridDimensions[2]){
```
skalarproduktraum, line 120: Flat indexing might work better here 👍

src/main/kotlin/graphics/scenery/DFTParser.kt
```diff
@@ -0,0 +1,132 @@
...
+            for (y in 0 until gridDimensions[1]){
+                for (x in 0 until gridDimensions[0]){
+                    val value = (((electronicDensity[x][y][z] - minDensity) / (maxDensity - minDensity)) * 255.0f).toInt()
+                    electronicDensityUInt.put(value.toByte())
```
skalarproduktraum, line 124: If `electronDensityUInt` is supposed to be exposed publicly (as it is currently), I'd suggest working on a `duplicate()` of it. Otherwise using `put` will always move around the `position` of the ByteBuffer and will lead to nasty issues if someone else is accessing it while you're writing 👍

skalarproduktraum: Oh and also, is it really UInt and not UShort or UByte?

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTExample.kt
```diff
@@ -0,0 +1,101 @@
...
+import kotlin.concurrent.thread
+import kotlin.math.pow
+
+/**
+ * <Description>
+ *  Visualizes a DFT snapshot.
+ *
+ * @author Lenz Fiedler <l.fiedler@hzdr.de>
+ */
+class DFTExample : SceneryBase("DFTExample", wantREPL = System.getProperty("scenery.master", "false").toBoolean()) {
```
skalarproduktraum, line 23: You can remove the wantREPL stuff here, I had unfortunately forgotten to remove it from TexturedCubeExample after a recent change, sorry for the confusion.

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTExample.kt
```diff
@@ -0,0 +1,101 @@
...
+            s.position = (snapshot.atomicPositions[i].mul(scalingFactor))
+            s.material.metallic = 0.3f
+            s.material.roughness = 0.9f
+            scene.addChild(s)
+        }
+
+        // Visualize the density data.
+
+        val volume = Volume.fromBuffer(emptyList(), snapshot.gridDimensions[0], snapshot.gridDimensions[1],
+                                        snapshot.gridDimensions[2], UnsignedByteType(), hub)
```
skalarproduktraum, line 48: Now I feel smart because I predicted your electron density is not actually UInt but UByte 🤣

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTExample.kt
```diff
@@ -0,0 +1,101 @@
...
+        volume.pixelToWorldRatio = snapshot.gridSpacings[0]*scalingFactor
+
+        volume.transferFunction = TransferFunction.ramp(0.0f, 0.3f, 0.5f)
+        scene.addChild(volume)
+        volume.addTimepoint("t-0", snapshot.electronicDensityUInt)
+        volume.goToLastTimepoint()
+
+
+        // One light in every corner.
+        val lights = (0 until 8).map {
```
skalarproduktraum, line 66: Lights do currently not influence the volume rendering (sorry). If you want better performance, just one light might be enough. You can also use `Light.createLightTetrahedron` which creates nicely positioned light sources for your scene, illuminating it from all directions 👍

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTExample.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        val cam: Camera = DetachedHeadCamera()
+        with(cam) {
+            position = Vector3f(0.0f, 0.0f, 5.0f)
+            perspectiveCamera(50.0f, 512, 512)
+
+            scene.addChild(this)
+        }
+
+        thread {
```
skalarproduktraum, line 86: This can be removed as it's empty.

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
```diff
@@ -0,0 +1,115 @@
...
+            snapshot.unitCellOrigin[2]).mul(scalingFactor)).add(
+            Vector3f(2.0f, 2.0f, 2.0f))
+        volume.colormap = Colormap.get("viridis")
+        volume.pixelToWorldRatio = snapshot.gridSpacings[0]*scalingFactor
+        volume.transferFunction = TransferFunction.ramp(0.0f, 0.3f, 0.5f)
+        scene.addChild(volume)
+
+
+        // One light in every corner.
+        val lights = (0 until 8).map {
```
skalarproduktraum, line 60: Same as the above comment for the lights.

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
```diff
@@ -0,0 +1,115 @@
...
+                    atoms[i].position = (snapshot.atomicPositions[i]).mul(scalingFactor)
+                }
+
+                volume.addTimepoint("t-${count}", snapshot.electronicDensityUInt)
+                volume.goToLastTimepoint()
+                volume.purgeFirst(10, 10)
+                Thread.sleep(500)
+                currentSnapshot++
+                count++
+                if (currentSnapshot == maxSnapshot)
```
skalarproduktraum, line 100: Could be simplified with module, e.g. `currentSnapshot = (currentSnapshot + 1) % maxSnapshot`

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
```diff
@@ -0,0 +1,115 @@
...
+        var currentSnapshot = 0
+        val maxSnapshot = 10
+        var count = 0
+        thread {
+            while (running) {
+                // Read new MD snapshot.
+                snapshot.parseCube(getDemoFilesPath() + "/volumes/dft/Fe_snapshot${currentSnapshot}_dens.cube")
+
+                // Visualize the atoms.
+                for(i in 0 until snapshot.numberOfAtoms) {
```
skalarproduktraum, line 89: I'd use `atoms.forEach { ... }` here instead, as it prevents nasty off-by-one bugs 👍

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
```diff
@@ -0,0 +1,115 @@
...
+                    // Shift the positions since the positions from the cube file are centers.
+                    atoms[i].position = (snapshot.atomicPositions[i]).mul(scalingFactor)
```
skalarproduktraum, line 91: Instead of `mul` you can just use the `*` operator

src/main/kotlin/graphics/scenery/DFTParser.kt
```diff
@@ -0,0 +1,132 @@
...
+                            }
+                        }
+                    }
+                }
+            }
+            counter++
+        }
+        // Converting to byte buffer.
+        counter = 0
+        electronicDensityUInt =  MemoryUtil.memAlloc((gridDimensions[0]*
```
skalarproduktraum, line 115: `memAlloc` creates off-JVM-heap memory, so that memory at some point needs to be released again. Best if you class implements the `AutoCloseable` interface, and you `memFree` the memory in the `close()` routine 👍

src/main/kotlin/graphics/scenery/DFTParser.kt
```diff
@@ -0,0 +1,132 @@
...
+                    {
+                        unitCellDimensions[0] = (gridSpacings[0]*gridDimensions[0])+unitCellOrigin[0]
+                        unitCellDimensions[1] = (gridSpacings[1]*gridDimensions[1])+unitCellOrigin[1]
+                        unitCellDimensions[2] = (gridSpacings[2]*gridDimensions[2])+unitCellOrigin[2]
+                        electronicDensity = Array(gridDimensions[0], { Array(gridDimensions[1],
+                            { FloatArray(gridDimensions[2]) } ) } )
+                    }
+                    // Parsing atomic positions.
+                    if (counter < 6+numberOfAtoms){
+                        atomicPositions[counter-6] = Vector3f((line.trim().split("\\s+".toRegex())[2]).toFloat(),
```
skalarproduktraum, line 79: Regex splitting is rather slow, if you're not splitting on super-special stuff, just consider using whitespace or sth.

skalarproduktraum: Also, you are repeatedly using `line.trim.split()` -- that's most likely your bottleneck, as I am not sure if the compiler has enough information available in order to optimise it out. I'd just split once per-line if possible.

src/main/kotlin/graphics/scenery/DFTParser.kt
```diff
@@ -0,0 +1,132 @@
...
+    var numberOfAtoms:Int = 0
+    /** Distance between two grid points in either direction in Bohr.*/
+    var gridSpacings = FloatArray(3) { 0.0f }
+    /** Number of gridpoints in 3D grid.*/
+    var gridDimensions = IntArray(3) { 0 }
+    /** Positions of the atoms in Bohr.*/
+    var atomicPositions = Array(0){ Vector3f()}
+    private var electronicDensity = Array(0, {Array(0, { FloatArray(0) } ) } )
+    /** Electronic density as scaled bytes, in scaled(1/Bohr).*/
+    lateinit var electronicDensityUInt : ByteBuffer
```
skalarproduktraum, line 28: Minor thing, but this is public now, you might wanna declare it as 
```kotlin
lateinit var electronicDensityUInt: ByteBuffer
    protected set
```

... which means you can get it public, but not reassign it 👍

src/main/kotlin/graphics/scenery/DFTParser.kt
```diff
@@ -0,0 +1,132 @@
...
+Parses a density functional theory (common simulation method in solid state physics and theoretical chemistry)
+calculation. Can be used to visualize single DFT calculations or DFT-MD (=DFT molecular dynamics simulations).
+
+ * @property[normalizeDensityTo] Defines to which value the density is scaled. This is useful when visualizing more then one DFT calculation at the
+    same time, in order to keep the density visualization consistent. Negative values mean the density is scaled
+    per snapshot. Default is -1.0f.
+ */
+class DFTParser (val normalizeDensityTo: Float = -1.0f){
+    /**  Number of Atoms.*/
+    var numberOfAtoms:Int = 0
```
skalarproduktraum, line 19: Space missing before type decl.

src/main/kotlin/graphics/scenery/DFTParser.kt
```diff
@@ -0,0 +1,132 @@
...
+    /** Distance between two grid points in either direction in Bohr.*/
+    var gridSpacings = FloatArray(3) { 0.0f }
```
skalarproduktraum, line 21: `floatArrayOf(0.0f, 0.0f, 0.0f)` is faster, as it won't incur the lambda overhead for each element you are initialising 👍

skalarproduktraum: Same goes for all the lines below 👍

skalarproduktraum (changes requested):
I have added a few remarks, some of them performance-related, some just style. Please let me know if you have any questions, I am happy to assist 👍

RandomDefaultUser:
Hey, thanks for all the helpful feedback! I'll have a look shortly.

RandomDefaultUser:
I just rebased the branch to be up-to-date with the  newest API and will look at your suggestions soon.

src/main/kotlin/graphics/scenery/DFTParser.kt
RandomDefaultUser: You're absolutely right, this is a Byte array. I have changed the name accordingly. I have also implemented the usage of a `duplicate` (I hope I understood correctly what you meant).

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTExample.kt
RandomDefaultUser: :smile: Yes, you're absolutely right there!

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTExample.kt
RandomDefaultUser: This was not about illuminating the volume, but about illuminating all the atoms from all sides. Will `Light.createLightTetrahedron` positively affect perfromance as well or does that only simply the call itself?

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTExample.kt
RandomDefaultUser: Done!

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTExample.kt
RandomDefaultUser: Right, good point :sweat_smile:

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
RandomDefaultUser: Good point, I changed that!

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
RandomDefaultUser: When I try to do that kotlin gives me an error, stating that it cannot resolve the reference. I presumed the problem was due to the different data types of float and vector and found that with `mul( )` I could make it work.

src/main/kotlin/graphics/scenery/DFTParser.kt
RandomDefaultUser: The problem is that .cube files are not necessarily consistent in their usage of whitespaces:

> In particular, the Gaussian specification is ambiguous about whitespace requirements, so parsing of CUBE files SHOULD accommodate some variation in the format, including (i) variable amounts/types of whitespace between the values on a given line, and (ii) the presence of leading and/or trailing whitespace on a given line.

https://h5cube-spec.readthedocs.io/en/latest/cubeformat.html#overview

Which is why I went with the Regex splitting. But apart from that you're absolutely right -  I was doing that way too often!

src/main/kotlin/graphics/scenery/DFTParser.kt
RandomDefaultUser: Ah yes, good idea, done!

src/main/kotlin/graphics/scenery/DFTParser.kt
RandomDefaultUser: Fixed!

src/main/kotlin/graphics/scenery/DFTParser.kt
RandomDefaultUser: I fixed that for the arrays I'm initializing, although for `electronicDensity` I cannot really do that since I am allocating it at runtime, or am I missing something here?

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTExample.kt
RandomDefaultUser: I applied around with LightTetraheadron a bit, but for my data, it always lead to one/multiple atom(s) directly in the corners always being partially illuminated. Is there a way to rotate it so I can but the light in corners?

src/main/kotlin/graphics/scenery/DFTParser.kt
RandomDefaultUser: Done!

RandomDefaultUser: (and I agree, that's a good idea. Using a 3D array was more to eliminate additional sources of error when coding up the parser, now that I understand what's going where, the switch makes sense).

src/main/kotlin/graphics/scenery/DFTParser.kt
RandomDefaultUser: Oh yes, good point, I totally forgot to do that... I just implemented it.

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
```diff
@@ -0,0 +1,117 @@
...
+        val snapshot = DFTParser()
+        snapshot.parseCube(getDemoFilesPath() + "/volumes/dft/Fe_snapshot0_dens.cube")
+
+        // Scales the DFT coordinates (which are in Bohr units) for a better VR experience.
+        val scalingFactor = 0.5f
+
+        // Visualize the atoms.
+        val atomicRadius = 0.5f*scalingFactor
+        val atoms: Array<Icosphere> = Array<Icosphere>(snapshot.numberOfAtoms) {Icosphere(atomicRadius, 4)}
+        atoms.zip(snapshot.atomicPositions).forEach {
```
skalarproduktraum, line 31: Considering this is an example: Would it make sense to introduce another class of type node, call it `AtomicSimulation` or something, and give it some factory methods to consume the cube files, and then directly add the volume data and atoms as child of itself?

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
```diff
@@ -0,0 +1,117 @@
...
+        val maxSnapshot = 10
+        var count = 0
+        thread {
+            while (running) {
+                // Read new MD snapshot.
+                snapshot.parseCube(getDemoFilesPath() + "/volumes/dft/Fe_snapshot${currentSnapshot}_dens.cube")
+
+                // Visualize the atoms.
+                atoms.zip(snapshot.atomicPositions).forEach {
+                    it.component1().spatial().position = it.component2().mul(scalingFactor)
```
skalarproduktraum, line 93: This is done repeatedly in the code and could probably be packaged in a function. Also, please replace `mul` with the `*` operator, IntelliJ will offer to import it 👍

skalarproduktraum (changes requested):
Two more things, I'd be very open to also discuss them interactively :-)

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
RandomDefaultUser: I think that's a great idea. That would also be helpful in the future. There's a plethora of different files used in computational physics and chemistry. This would also make it easier to only read atoms; sometimes one is not interested in volumetric data, or it is simply not present. 
Should I convert this PR to  draft and start working on something like this? We can also chat about this on Mattermost or via Zoom or something.

src/test/kotlin/graphics/scenery/tests/examples/basic/DFTMDExample.kt
RandomDefaultUser: Done!

RandomDefaultUser:
As discussed with @skalarproduktraum I am finishing up this PR by

- [ ] Moving the source to a package
- [ ] integrating the test files in the demo folder
- [ ] Adding a "rootFolder" option for the AtomicSimulation class
- [ ] Making docstrings consistent with scenery
- [ ] Making the coding style consistent with scenery

PR #444, opened 2021-10-07
Correct VolumeManager teardown

skalarproduktraum:
This PR improves the teardown behaviour of `VolumeManager`, such that correct reinitialisation, e.g. via `SceneryBase.replaceRenderer()` is possible.

It also introduces a new `close()` method for `Renderable`.

With this, an issue reported by @RuoshanLan is fixed, where volumes would not correctly show after the renderer has been restarted without an application restart.

This PR also fixes an issue where wrong near/far planes would be used in `DetachedHeadCamera` to get a projection matrix from a `Display` interface.


RuoshanLan:
I just checked the branch again, but the issue is still there

https://user-images.githubusercontent.com/10786204/140912073-2f358962-b1c7-477f-9dde-937093220c47.mp4

skalarproduktraum:
This PR now also fixes the issue @RuoshanLan reported above, where different eyes had different z-ordering. This was a regression introduced in #406.

PR #446, opened 2021-10-13
Rendering particles as glyphs

PowerOfNames:
Added the functionality to render spherical particle glyphs with help of geometry shader fabricated billboards and fragment shader based ray-casting.

Files added: 

- ParticleGlyphs.kt (src/main/kotlin/primitives/)
- ParticleGlyphs.vert
- PartcileGlyphs.geom
- PartcleGlyphs.frag
- ParticleGlyphsForward.frag (srcmain/resources/backends/shaders/)
- ParticleGlyphsExample.kt (src/test/kotlin/graphics.scenery.tests/examples.basic/)


skalarproduktraum:
@PowerOfNames What's the status here? Did you factor in all requested changes?

src/main/kotlin/graphics/scenery/primitives/ParticleGlyphs.kt
```diff
@@ -28,20 +36,17 @@ class ParticleGlyphs @JvmOverloads constructor(var positions: List<Vector3f>, va
...
         geometry{
-            vertices.position(0)
-            vertices.limit(positions.size * 3)
-            positions.forEach { v ->
-                v.get(vertices)
-                vertices.position(vertices.position() + 3)
-            }
+
+            vertices = BufferUtils.allocateFloat(positions.limit())
+            vertices.put(positions)
```
skalarproduktraum, line 41: This is allocating native memory, so your class needs a `close()` method to free that memory again 👍

src/main/kotlin/graphics/scenery/primitives/ParticleGlyphs.kt
```diff
@@ -56,8 +61,9 @@ class ParticleGlyphs @JvmOverloads constructor(var positions: List<Vector3f>, va
 
     override fun createGeometry(): Geometry {
         return object: DefaultGeometry(this) {
-            override var vertices: FloatBuffer = BufferUtils.allocateFloat(positions.size * 3)
-            override var normals: FloatBuffer = BufferUtils.allocateFloat(properties.size * 3)
+            override var vertices: FloatBuffer = BufferUtils.allocateFloat(positions.limit())
```
skalarproduktraum, line 64: That should not be `positions.limit()`, but `positions.remaining()`

src/main/kotlin/graphics/scenery/primitives/ParticleGlyphs.kt
```diff
@@ -56,8 +61,9 @@ class ParticleGlyphs @JvmOverloads constructor(var positions: List<Vector3f>, va
...
+            override var texcoords: FloatBuffer = BufferUtils.allocateFloat(properties.limit())
```
skalarproduktraum, line 65: See above

src/main/kotlin/graphics/scenery/primitives/ParticleGlyphs.kt
```diff
@@ -56,8 +61,9 @@ class ParticleGlyphs @JvmOverloads constructor(var positions: List<Vector3f>, va
...
+            override var normals: FloatBuffer = BufferUtils.allocateFloat(colors.limit())
```
skalarproduktraum, line 66: See above

src/main/kotlin/graphics/scenery/primitives/ParticleGlyphs.kt
```diff
@@ -92,4 +98,21 @@ class ParticleGlyphs @JvmOverloads constructor(var positions: List<Vector3f>, va
 
         return newMaterial
     }
+
+    fun updatePositions(positions: FloatBuffer)
+    {
+        geometry{
+            vertices = positions
```
skalarproduktraum, line 105: Careful here. In the constructor, you copy the data from the positions object you hand over to vertices, here, you only change the reference. This might come back and bite you.

src/main/resources/graphics/scenery/backends/shaders/ParticleGlyphs.frag
```diff
@@ -1,25 +1,22 @@
...
 #extension GL_ARB_separate_shader_objects: enable
 
-layout(location = 0) in SilhouetteData
-{
+layout(location = 0) in SilhouetteData {
     vec3 Position;
     vec2 TexCoord;
     flat vec3 Center;
-    flat vec3 Properties;
+    flat vec2 Properties;
```
skalarproduktraum, line 8: Just checking - these are intended to be `flat` and not interpolated?

skalarproduktraum (changes requested):

src/main/resources/graphics/scenery/backends/shaders/ParticleGlyphs.frag
PowerOfNames: Yes, no interpolation wanted here!

src/main/kotlin/graphics/scenery/primitives/ParticleGlyphs.kt
PowerOfNames: Okay, so I need to get the reference here. thought kotlin is working with references all the time, if not explicitly said to do differently?

src/main/kotlin/graphics/scenery/primitives/ParticleGlyphs.kt
PowerOfNames: Where do I need to call close then?

src/main/kotlin/graphics/scenery/primitives/ParticleGlyphs.kt
PowerOfNames: And why is that? remaining is the number between the current pointer and the limit. Shouldn't the pointer alsways be on the first position?

PR #471, opened 2022-01-28
Synchronization 2.0

moreApi:
# Network Documentation

The network synchronization of scenery has two targets. First for CAVE setups the scene needs to be rendered by multiple nodes from different perspectives and second multi user volume viewing sessions using networked computers.

## How to Start
The networking capabilities of scenery are enabled via VM parameters. The server has to be started with 
`-Dscenery.Server=true` 
and the client needs to be pointed to the server via
`-Dscenery.ServerAddress=tcp://127.0.0.1`

The server can be pretty much any scenery scene (maybe with some adjusments, see below). For a simple client use [SlimClient].

#### Additional VM Parameters
- For server and client
	- `scenery.MainPort` - Port for the main channel. Default: `6040`
	- `scenery.BackchannelPort` - Port for the Backchannel. Default: `6041`

## Relevant Examples
- [SimpleNetworkExample]
- [NetworkVolumeExample]
- [SlimClient]

## How it works
Once the server application is started it begins scanning and registering the scene graph for objects which implement the [Networkable] interface. By default that are all nodes connected to the root node and their attributes. 

When a client requests a resync all registered objects are serialized and send to all connected clients. Upon receiving an object the client checks whether it has seen one with that id before and if not adds its to its scene. If the objects parent has not been synced yet it is put in a waiting queue and added once the parent has been synced it is added. If the object is already there the existing object is simply updated with the values of the new object.

If the server registers a change in a registered object it sends an update with the new object to all clients.

#### Relevant Classes
- [Networkable]
- [NodePublisher]
- [NodeSubscriber]

## Slim Client
The [SlimClient] is a scenery application which offers a mostly bare scene in which scenes from a server might be loaded. In addition to that it offers its own camera if desired. If the VM parameter 
`-Dscenery.RemoteCamera=true`
is set cameras from the server scene are disregarded and a local one is used. If it is not or to false set the server scene has to provide a camera.

## Network Adjustments
- Camera
	- [Camera.wantsSync] - allows disabling the registration of the camera for syncing. Default: true
- Material
	- [DefaultMaterial.syncronizeTextures] - if set to true trys to transmit the textures over the network. Should be disabled for large textures. Default: True
- Volume
	- [Volume.forNetwork] - Volume creation method for networkable volumes. For more see section below.

### New Objects
There are two ways to add objects for the client. The default way is to take the deserialized object from the server and simply add it to its corresponding parent. This works for most cases. But some objects require the constructor be run locally or with specific parameters. For those cases there are the `getConstructorParameters()` and `constructWithParameters(parameters)` functions in the `Networkable` interface.  No matter how the object was created afterwards the update method is called with the same object  from the server again. The update happens after the object has been added to the scene graph.   

#### Example
[PointLight]

### Updating Objects
Every `Networkable` object needs to implement the `update` method. In the `update` method a fresh copy from the server is given as parameter and should be used to copy over relevant values to the client-side object.

Properties which are marked `@Transient` are not serialized and therefore not available on the client-side copy of a object from the server. If those properties need to be synced anyway they should be returned in a serializable form in an overwrite of the `getAdditionalUpdateData()` method. This method is called on the server at the serialization and the results are  transmitted next to the object. The transmitted result of the `getAdditionalUpdateData()` function is another parameter of the update function. Special attention has to be paid to parent classes which might also have additional data. This data has to be handled manually in the overriding methods.  

References to other `Networkable` objects besides parent/child relations of the scene graph and node/attribute relations can't be synced automatically. For those the `Networkable.networkID` needs to be saved as additional data on the server side and resolved in the `update` method with the `getNetworkable` lambda parameter. 

#### Attention: The first Update  
If the object was not created with a local constructor but deserialization then the first update will be "with itself". This might be a source of sneaky bugs when for example lists which where cleared on the "client" object are now also empty on the "server" object. 

#### Examples
[DefaultMaterial] - AdditionalUpdateData
[DefaultSpatial] - GetNetworkable, First Update

### Making Changes Known
For the server to register a change the `modifiedAt` property need to be updated. For convenience the `updateModifiedAt` method simply may be called like in [DefaultSpatial].

### Disabling Syncing
It might be needed to prevent the server from registering an object. If `wantsSync` returns false the object will be skipped in the registration like in [Camera].  

### Registering Subcomponents
For the server to register objects out of the scene graph they need to be returned by `getSubcomponents`. The registration of attributes is handled that way for [DefaultNode]. One could extends this for rarely changing data objects to reduce the amount of times they are transmitted. 

### On transmitting lambdas
At the time of writing serializing lambdas is wonky. Don't expect it to work. An alternative is to generate them new on the client side in the `constructWithParameters` or `update` method.

## Volumes
Syncing volume data is currently not possible. Therefore the data has to be available  locally. 

To initialize a volume node with sync support the `Volume.forNetwork` create method should be used. It takes an implementation of the [VolumeInitalizer] interface.

There are already two implementation in this repo. The first is part of scenery it self [VolumeFileSource].

`VolumeFileSource` has two parameters with each two options:
- `path`
	- `Given(val filePath: String)` - for fixed file path witch are the same on every machine (eg. network drive or something like "C://Volume")
	- `Settings(val settingsName: String = "VolumeFile")` - the file path is taken from the VM parameter "-DVolumeFile=$path$" of each individual application
	- `Resource(val path:String)` - the volume is a resource reachable by the java loader
- `type`
	- `TIFF` - tiff file format
	- `SPIM` - Spim xml data format

The other implementation is [IJVolumeInitializer] which can be found in the [NetworkVolumeExample]. It takes a path/url and opens it using the ImageJ framework.

#### Examples

[NetworkVolumeExample]


src/main/kotlin/graphics/scenery/Box.kt
```diff
@@ -135,6 +137,20 @@ open class Box @JvmOverloads constructor(val sizes: Vector3f = Vector3f(1.0f, 1.
         boundingBox = generateBoundingBox()
     }
 
+    override fun getConstructorParameters(): Any? {
```
skalarproduktraum, line 140: Shouldn't these two methods be static, aka, be moved to the companion object, as they are essentially factory methods?

src/main/kotlin/graphics/scenery/DetachedHeadCamera.kt
```diff
@@ -1,13 +1,13 @@
 package graphics.scenery
 
-import org.joml.Matrix4f
-import org.joml.Vector3f
 import com.jogamp.opengl.math.Quaternion
```
skalarproduktraum, line 3: This import should be removed.

src/main/kotlin/graphics/scenery/DetachedHeadCamera.kt
```diff
@@ -106,47 +106,52 @@ class DetachedHeadCamera(@Transient var tracker: TrackerInput? = null) : Camera(
...
 //        val hr = Matrix4f.fromQuaternion(this.headOrientation)
-                return tracker?.getWorkingTracker()?.getPose()?.times(tr) ?: Matrix4f().set(this.rotation) * tr
-            }
-            /**
-             * Returns this camera's transformation for eye with index [eye], taking an eventually existing [TrackerInput]
-             * into consideration as well.
-             */
-            override fun getTransformationForEye(eye: Int): Matrix4f {
-                val tr = Matrix4f().translate(this.position * (-1.0f))
+            return cam.tracker?.getWorkingTracker()?.getPose()?.times(tr) ?: Matrix4f().set(this.rotation) * tr
```
skalarproduktraum, line 133: This is an unwanted code style change

src/main/kotlin/graphics/scenery/DetachedHeadCamera.kt
```diff
@@ -106,47 +106,52 @@ class DetachedHeadCamera(@Transient var tracker: TrackerInput? = null) : Camera(
...
-                    super.projection = value
-                    field = value
-                }
-            /**
-             * Returns this camera's transformation matrix, taking an eventually existing [TrackerInput]
-             * into consideration as well.
-             */
-            override fun getTransformation(): Matrix4f {
-                val tr = Matrix4f().translate(this.position * (-1.0f))
+    override fun createSpatial(): CameraSpatial = DetachedHeadCameraSpatial(this)
```
skalarproduktraum, line 109: This is an unwanted code style change

src/main/kotlin/graphics/scenery/net/NetworkEvent.kt
```diff
@@ -0,0 +1,13 @@
+package graphics.scenery.net
+
```
skalarproduktraum, line 2: Kdoc comment with authors etc. missing

src/main/kotlin/graphics/scenery/net/NodeSubscriber.kt
```diff
@@ -1,117 +1,280 @@
...
-                        val oSpatial = o.spatialOrNull()
-                        if (oSpatial != null) {
-                            node.ifSpatial {
-                                position = oSpatial.position
-                                rotation = oSpatial.rotation
-                                scale = oSpatial.scale
-                            }
-                        }
-                        node.visible = o.visible
+    fun startListening() {
```
skalarproduktraum, line 60: KDoc missing

src/main/kotlin/graphics/scenery/net/NodeSubscriber.kt
```diff
@@ -1,117 +1,280 @@
...
 
+    private val addressSubscribe = "$ip:$portPublish"
+    private val addressBackchannel = "$ip:$portBackchannel"
+    var subscriber: ZMQ.Socket = context.createSocket(SocketType.SUB)
+    var backchannel: ZMQ.Socket = context.createSocket(SocketType.PUB)
+
+    private val networkObjects = hashMapOf<Int, NetworkWrapper<*>>()
+    private val eventQueue = LinkedBlockingQueue<NetworkEvent>()
+    private val waitingOnNetworkable = mutableMapOf<Int, List<Pair<NetworkEvent, WaitReason>>>()
+    private var listening = false
```
skalarproduktraum, line 42: This should probably be `@Synchronized` so it's the same on all threads.

src/main/kotlin/graphics/scenery/serialization/VolumeSerializer.kt
```diff
@@ -1,77 +0,0 @@
-package graphics.scenery.serialization
```
skalarproduktraum, line 1: Why did you delete this?

src/main/kotlin/graphics/scenery/volumes/OrthoView.kt
```diff
@@ -22,8 +23,9 @@ import kotlin.concurrent.thread
  * [Volume.slicingMode] should be set to [Volume.SlicingMode.None]
  *
  * @param key The input key for optional plane movement behavior. See Input Docs for format.
+ * @param hub only needed when [key] is set
  */
-fun createOrthoView(volume: Volume, key: String? = null) {
+fun createOrthoView(volume: Volume, key: String? = null, hub: Hub?) {
```
skalarproduktraum, line 28: Why does orthoview need a hub now?

src/main/kotlin/graphics/scenery/volumes/TransferFunction.kt
```diff
@@ -32,7 +32,7 @@ open class TransferFunction(val name: String = "") {
     @Transient protected val buffer: ByteBuffer = MemoryUtil.memCalloc(textureSize * 4 * textureHeight)
 
     /** Indicator whether the auxiliary texture needs to be reuploaded. */
-    var stale = true
+    @Transient var stale = true
```
skalarproduktraum, line 35: Should probably be solved differently...

src/main/kotlin/graphics/scenery/volumes/Volume.kt
```diff
@@ -32,16 +29,21 @@ import graphics.scenery.attribute.renderable.DelegatesRenderable
 import graphics.scenery.attribute.renderable.Renderable
 import graphics.scenery.attribute.spatial.DefaultSpatial
 import graphics.scenery.attribute.spatial.HasCustomSpatial
+import graphics.scenery.net.Networkable
 import graphics.scenery.numerics.OpenSimplexNoise
 import graphics.scenery.numerics.Random
 import graphics.scenery.utils.LazyLogger
 import graphics.scenery.volumes.Volume.VolumeDataSource.SpimDataMinimalSource
+import ij.IJ
```
skalarproduktraum, line 37: Nope nope nope nope :-D

src/main/kotlin/graphics/scenery/volumes/VolumeManager.kt
```diff
@@ -828,5 +828,29 @@ class VolumeManager(
     companion object {
         /** Static [ForkJoinPool] for fill task submission. */
         protected val forkJoinPool: ForkJoinPool = ForkJoinPool(max(1, Runtime.getRuntime().availableProcessors()))
+
+        fun regenerateVolumeManagerWithExtraVolume(volume: Volume, hub: Hub ) {
```
skalarproduktraum, line 832: This is probably the spot to do what you did in SceneryBase 👍

src/test/kotlin/graphics/scenery/tests/examples/volumes/CroppingExample.kt
```diff
@@ -69,10 +69,12 @@ class CroppingExample : SceneryBase("Volume Cropping example", 1280, 720) {
         origin.material().diffuse = Vector3f(0.8f, 0.0f, 0.0f)
         scene.addChild(origin)
 
-        val imp: ImagePlus = IJ.openImage("https://imagej.nih.gov/ij/images/t1-head.zip")
-        val img: Img<UnsignedShortType> = ImageJFunctions.wrapShort(imp)
+        val resource = Volume.VolumeFileSource(
+            Volume.VolumeFileSource.VolumePath.Resource("/graphics/scenery/tests/unit/volume/t1-head.zip"),
```
skalarproduktraum, line 73: The idea here is good, but this needs to work without the whole ImageJ dependency.

src/test/kotlin/graphics/scenery/tests/unit/network/PubSubTest.kt
```diff
@@ -0,0 +1,256 @@
...
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+
+/**
+ * Integration tests for [NodePublisher] and [NodeSubscriber]
+ */
+class PubSubTest {
```
skalarproduktraum, line 24: The test class name must reflect the class being tested.

src/test/kotlin/graphics/scenery/tests/unit/network/PublisherTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import graphics.scenery.net.NetworkWrapper
+import graphics.scenery.net.NodePublisher
+import org.joml.Vector3f
+import org.junit.Test
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+
+class PublisherTest {
```
skalarproduktraum, line 21: The test class name must reflect the class being tested.

src/test/kotlin/graphics/scenery/tests/unit/network/SubscriberTest.kt
```diff
@@ -0,0 +1,134 @@
...
+import graphics.scenery.Hub
+import graphics.scenery.Scene
+import graphics.scenery.net.*
+import org.junit.Test
+import java.io.ByteArrayInputStream
+import java.io.ByteArrayOutputStream
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+
+class SubscriberTest {
```
skalarproduktraum, line 16: The test class name must reflect the class being tested.

skalarproduktraum (changes requested):

src/main/kotlin/graphics/scenery/BoundingGrid.kt
```diff
@@ -28,18 +29,34 @@ open class BoundingGrid : Mesh("Bounding Grid") {
     /** Grid color for the bounding grid. */
     @ShaderProperty
     var gridColor: Vector3f = Vector3f(1.0f, 1.0f, 1.0f)
```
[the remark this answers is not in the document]
moreApi: sadly kryo hates delegates

src/main/kotlin/graphics/scenery/PointLight.kt
```diff
@@ -14,7 +15,7 @@ import org.joml.Vector4f
  * @author Ulrik Günther <hello@ulrik.is>
  * @constructor Creates a PointLight with default settings, e.g. white emission color.
  */
-class PointLight(radius: Float = 5.0f) : Light("PointLight") {
+class PointLight(val radius: Float = 5.0f) : Light("PointLight") {
```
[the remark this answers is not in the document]
moreApi: point light does not support a changable radius

src/main/kotlin/graphics/scenery/attribute/material/DefaultMaterial.kt
```diff
@@ -1,25 +1,51 @@
...
+        }
     override var metallic: Float = 0.0f
+        set(value) {
+            field = value
+            updateModifiedAt()
+        }
     override var blending: Blending = Blending()
-    @Volatile @Transient override var textures: TimestampedConcurrentHashMap<String, Texture> = TimestampedConcurrentHashMap()
+    @Volatile
+    @Transient
```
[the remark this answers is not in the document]
moreApi: If I remember correctly, I did not want to transmit potentially large textures and if someone wants to transmit textures they can use the custom sync stuff to do it themself.

build.gradle.kts
```diff
@@ -48,6 +48,8 @@ dependencies {
     implementation("org.scijava:scripting-javascript")
     implementation("org.scijava:scripting-jython")
     implementation("net.java.dev.jna:jna-platform:5.9.0")
+    implementation("net.imagej:imagej")
```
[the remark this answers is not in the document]
moreApi: removed and reworked volume initialization accordingly

src/main/kotlin/graphics/scenery/DetachedHeadCamera.kt
moreApi: Then we should open an issue for that. It was not added in this PR.

src/main/kotlin/graphics/scenery/attribute/material/DefaultMaterial.kt
```diff
@@ -1,25 +1,51 @@
...
     override var cullingMode: Material.CullingMode = Material.CullingMode.Back
     override var depthTest: Material.DepthTest = Material.DepthTest.LessEqual
     override var wireframe: Boolean = false
     override var timestamp: Long = System.nanoTime()
+    override var modifiedAt = Long.MIN_VALUE
+
+    var synchronizeTextures = true
 
     /** Companion object for Material, emulating static methods */
     companion object Factory {
```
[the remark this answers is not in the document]
moreApi: I dont know. FrauZufall wrote this ^^ Maybe not?

src/main/kotlin/graphics/scenery/attribute/spatial/DefaultSpatial.kt
```diff
@@ -10,39 +12,72 @@ import net.imglib2.RealLocalizable
...
-    override var world: Matrix4f by Delegates.observable(Matrix4f().identity()) { property, old, new -> propertyChanged(property, old, new) }
-    override var model: Matrix4f by Delegates.observable(Matrix4f().identity()) { property, old, new -> propertyChanged(property, old, new) }
-    override var view: Matrix4f by Delegates.observable(Matrix4f().identity()) { property, old, new -> propertyChanged(property, old, new) }
-    override var projection: Matrix4f by Delegates.observable(Matrix4f().identity()) { property, old, new -> propertyChanged(property, old, new) }
-    override var scale: Vector3f by Delegates.observable(Vector3f(1.0f, 1.0f, 1.0f)) { property, old, new -> propertyChanged(property, old, new) }
-    override var rotation: Quaternionf by Delegates.observable(Quaternionf(0.0f, 0.0f, 0.0f, 1.0f)) { property, old, new -> propertyChanged(property, old, new) }
-    override var position: Vector3f by Delegates.observable(Vector3f(0.0f, 0.0f, 0.0f)) { property, old, new -> propertyChanged(property, old, new) }
+open class DefaultSpatial(@Transient protected var node: Node = DefaultNode()) : Spatial {
+    override var world: Matrix4f = Matrix4f().identity()
```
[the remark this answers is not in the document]
moreApi: see converation on camera

src/main/kotlin/graphics/scenery/net/NetworkEvent.kt
moreApi: added

src/main/kotlin/graphics/scenery/net/NodeSubscriber.kt
moreApi: The hand-of point between the network thread and update/main-loop thread is the eventQueue which should be thread safe. The maps are ony accessed by the update thread and bools can't be @Synchronised

src/main/kotlin/graphics/scenery/volumes/OrthoView.kt
moreApi: To add a addOrthoViewDragBehavior. This is an optional convinience feature. Hub can be null if the feature is not desired. Previously the function got the hub from volume which loses its hub property in this PR.

src/main/kotlin/graphics/scenery/volumes/Volume.kt
moreApi: removed

src/test/kotlin/graphics/scenery/tests/examples/volumes/CroppingExample.kt
moreApi: thanks. resolved

src/main/kotlin/graphics/scenery/DetachedHeadCamera.kt
```diff
@@ -24,7 +24,7 @@ import kotlin.reflect.KProperty
 class DetachedHeadCamera(@Transient var tracker: TrackerInput? = null) : Camera() {
 
     override var width: Int = 0
-        get() = if(tracker != null && tracker is Display && tracker?.initializedAndWorking() == true) {
+        get() = if (tracker != null && tracker is Display && tracker?.initializedAndWorking() == true) {
```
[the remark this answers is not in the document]
moreApi: Do we want a code style that is different from the default Intellij IDEA style abweicht?

If yes that would mean one always needs to manualy check each file after a automatic formatting.

src/main/kotlin/graphics/scenery/Camera.kt
```diff
@@ -45,24 +46,51 @@ open class Camera : DefaultNode("Camera"), HasRenderable, HasMaterial, HasCustom
     var right: Vector3f = Vector3f(1.0f, 0.0f, 0.0f)
     /** FOV of the camera **/
     open var fov: Float = 70.0f
+        set(v) {
```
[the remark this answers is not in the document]
skalarproduktraum: Try adding the `@delegate:Transient` annotation 👍

src/main/kotlin/graphics/scenery/DetachedHeadCamera.kt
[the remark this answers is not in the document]
skalarproduktraum: There was actually a code style file at some point. IntelliJ's defaults have also changed over time. I'll see that we can resolve this, but please remove the code style changes for this PR, as this really distracts from the actual changes 👍

src/main/kotlin/graphics/scenery/attribute/material/DefaultMaterial.kt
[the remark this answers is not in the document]
skalarproduktraum: Okay, worth a shot. Let's see how this behaves.

src/main/kotlin/graphics/scenery/attribute/material/DefaultMaterial.kt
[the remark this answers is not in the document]
skalarproduktraum: Okay, then leave it for the moment.

src/main/kotlin/graphics/scenery/SceneryBase.kt
```diff
@@ -172,46 +174,34 @@ open class SceneryBase @JvmOverloads constructor(var applicationName: String,
...
-
-                hub.add(SceneryElement.NodeSubscriber, subscriber)
-                scene.discover(scene, { true }).forEachIndexed { index, node ->
-                    subscriber.nodes.put(index, node)
-                }
-
-                while (running && !shouldClose) {
-                    subscriber.process()
-                    Thread.sleep(2)
+        scene.onChildrenAdded["VolumeManager"] = { _,node ->
```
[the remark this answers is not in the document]
skalarproduktraum: The volume manager knows about the volumes it manages, once it's created, it can get a reference to the scene, and add the hook. Alternatively, you could also put this hook e.g. in the preDraw() routine of VolumeManager.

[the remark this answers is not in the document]
moreApi: When and where is the VolumeManager created the first time?

src/main/kotlin/graphics/scenery/volumes/OrthoView.kt
skalarproduktraum: Alright for the moment, we'd probably have to think of a better way down the road.

src/main/kotlin/graphics/scenery/volumes/VolumeManager.kt
skalarproduktraum: Yeah that was not correct, the constructor of `Volume` should actually be the correct place as mentioned above 👍

src/main/kotlin/graphics/scenery/DetachedHeadCamera.kt
moreApi: Nope, this is needed. Kryo cant sync anonymous classes

src/main/kotlin/graphics/scenery/Camera.kt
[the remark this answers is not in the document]
moreApi: That applyed to world and position of spatial gets us:
`Exception in thread "main" java.lang.NullPointerException
	at graphics.scenery.attribute.spatial.Spatial$DefaultImpls.propertyChanged$default(Spatial.kt:114)
	at graphics.scenery.attribute.spatial.DefaultSpatial$special$$inlined$observable$2.afterChange(Delegates.kt:71)
	at kotlin.properties.ObservableProperty.setValue(ObservableProperty.kt:41)
	at graphics.scenery.attribute.spatial.DefaultSpatial.setRotation(DefaultSpatial.kt:45)
	at graphics.scenery.attribute.spatial.DefaultSpatial.update(DefaultSpatial.kt:428)
	at graphics.scenery.net.NodeSubscriber.processUpdateEvent(NodeSubscriber.kt:157)
	at graphics.scenery.net.NodeSubscriber.processUpdateEvent(NodeSubscriber.kt:227)
	at graphics.scenery.net.NodeSubscriber.networkUpdate(NodeSubscriber.kt:113)
	at graphics.scenery.SceneryBase$sceneryMain$3.invoke(SceneryBase.kt:186)`

which looks to me like a black magic kryo error and I neither have the time and frog eyes to dispell it.

PR #474, opened 2022-03-15
Volume: Improved raycast behaviour

skalarproduktraum:
This PR fixes the volume ray caster to only sample from volumes within their own bounding box.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -88,7 +88,7 @@ dependencies {
     api("sc.fiji:bigdataviewer-vistools:1.0.0-beta-28")
 
     //TODO revert to official BVV
-    api("graphics.scenery:bigvolumeviewer:a6b021d")
+    api("graphics.scenery:bigvolumeviewer:7f36eda")
 
     implementation("com.github.LWJGLX:lwjgl3-awt:cfd741a6")
     implementation("org.janelia.saalfeldlab:n5")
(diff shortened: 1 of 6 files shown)


skalarproduktraum:
Did you also add the `localNear` and `localFar` keys in the place where you've created the VolumeManager? This is also not (yet) done in the CustomVolumeManagerExample.

skalarproduktraum:
Is this good to merge, @aryaman-gupta?

skalarproduktraum:
Ping again @aryaman-gupta

aryaman-gupta:
@skalarproduktraum sorry for the delay in replying

Unfortunately, when a volume is sliced into multiple chunks, some artefacts are visible at the borders. It appears as though a single slice of voxels is not sampled correctly. The attached image is an example. The image shows the Stagbeetle dataset (Meister Eduard Gröller, Georg Glaeser, Johannes Kastner, 2005)

DistributedVolumeRenderer - 2022-05-25 00 02 31

To reproduce, the easiest way may be to run the `VolumeFromFileExample` of the `scenery-insitu` repository. Before running the application, the Stagbeetle dataset would need to be divided into two parts and placed in directories named Part1 and Part2, each with its own `stacks.info` file indicating the dimensions of the respective part.

skalarproduktraum:
Superseded by #622.

PR #476, opened 2022-04-14
Vulkan: macOS and Apple Silicon support

skalarproduktraum:
This PR adds support for using Vulkan on macOS, especially on Apple Silicon chips.

Before merging the following needs to work:
- [X] Headless rendering
- [X] Rendering using the SwingSwapchain for embedding, plus REPL
- [x] Rendering using ~~GLFW windows~~ switched to Swing for macOS, plus REPL
- [X] Tested on macOS/ARM64
- [ ] Tested on macOS/AMD64
- [x] https://github.com/LWJGLX/lwjgl3-awt/issues/37 needs to be addressed and a new artifact including ARM64 support available
- [x] CI passing
- [x] All important integration tests running

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,8 +11,6 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            java: 8
-          - os: ubuntu-latest
             java: 11
           - os: ubuntu-latest
             java: 16
@@ -42,8 +40,6 @@ jobs:
       matrix:
         include:
           - os: windows-latest
-            java: 8
-          - os: windows-latest
             java: 11
           - os: windows-latest
             java: 16
@@ -68,8 +64,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: macos-latest
-            java: 8
           - os: macos-latest
             java: 11
           - os: macos-latest
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -76,7 +76,7 @@ unit-tests-no-gpu:
       - "hs_err_*"
 
 scenery-nvidia:
-  image: scenerygraphics/nvidia-vulkan:1.2.170.0-ubuntu20.04
+  image: scenerygraphics/nvidia-vulkan:1.3.216.0-ubuntu20.04
   <<: *base-job-gpu
   after_script:
     - nvidia-smi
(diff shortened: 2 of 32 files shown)


mkitti:
Regarding the JHDF5 fat jar, let me know how I can help with this. The far jar is convenient but perhaps is not the most modular.

I'm also still trying to work with the original authors on their support, but it is possible to we may need to consider a formal fork.

skalarproduktraum:
Thanks for the offer @mkitti - and actually also for making this work on M1! At the moment I am pulling it just from your Github release, which works for the moment. It's not ginormous in size and packs all the relevant architectures, so that's fine. If you'd have this at some point available in a Maven repo, that'd be a plus, but no rush 👍

mkitti:
I just wanted point out that the cisd.base artifact is in maven:
https://mvnrepository.com/artifact/cisd/base/18.09.0

JHDF5 19.04 is also available on maven via
https://mvnrepository.com/artifact/cisd/jhdf5/19.04.0

They currently use POMs located here: https://github.com/scijava/pom-scijava/tree/master/non-maven

However, that version of JHFD5 does not include the M1 binaries or Blosc support.

If we could pre-extract the binaries somewhere, we could point the currently mavenized package to M1 binaries as described here:
https://unlimited.ethz.ch/display/JHDF/JHDF5+FAQ#JHDF5FAQ-Whataretheoptionstoprovidethenativelibraries?

Essentially we could set the `-Dnative.libpath.jhdf5` to the location of the JHDF5 shared library for the M1.

skalarproduktraum:
@mkitti, I had missed your comment before merging this - and I like the idea! How about publishing an artifact just for M1 to the SciJava Maven for the moment? That could then be used as a regular dependency and extracted if needed. 

On a related note, it looks as if the JHDF5 people also just added M1 support: https://sissource.ethz.ch/sispub/jhdf5/-/commit/ecc1cfb7a1f8dab428906c874e3e7b407b450f77

mkitti:
Yes, I am in communication with Bernd Rinn about this issue. As you can see, they have pulled the Janelia aarch64 branch into their repository.

Note that the jar you are currently using goes further than what they have presently integrated. In particular, you have downloaded a branch that includes Blosc compression support.

I expect that ETH SIS will issue an updated JHDF5 release. Next, I will coordinate with @ctrueden about distributing that release via maven.

I will probably release accessory packages to JHDF5 that include
1) Split HDF5 and JHDF5 (JNI bindings) shared libraries.
2) Compression plugin support

PR #485, opened 2022-09-05
Raycasting fix

PowerOfNames:
Added SceneRayCastingExample.kt, which contains a simple Scene wioth a Plane, a camera and the ability to hit key '1' to shoot a ray and debug print the hit result.

Changed Camera.kt's getNodesFromScreenSpace() to also take in an optional boolean to choose whether the ray origin should be on the worldSpace screen (starts ray at the position the screen would be when it is located in front f the camera and the ray 'shoots' through it into the scene) or the camera origin.

screenPointToRay now used the prior boolean (camOrigin) to either return a pair of worldPos and worldDir (boolean false) or origin (camera) and worldDir (boolean true)

### DIFF
--- a/src/main/kotlin/graphics/scenery/Camera.kt
+++ b/src/main/kotlin/graphics/scenery/Camera.kt
@@ -194,11 +194,12 @@ open class Camera : DefaultNode("Camera"), HasRenderable, HasMaterial, HasCustom
     fun getNodesForScreenSpacePosition(
         x: Int, y: Int,
         ignoredObjects: List<Class<*>> = listOf<Class<*>>(BoundingGrid::class.java),
+        camOrigin: Boolean = false,
         debug: Boolean = false
     ): Scene.RaycastResult {
         return getNodesForScreenSpacePosition(x, y, { n: Node ->
             !ignoredObjects.any { it.isAssignableFrom(n.javaClass) }
-        }, debug)
+        }, camOrigin, debug)
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/main/kotlin/graphics/scenery/Camera.kt
```diff
@@ -227,7 +229,7 @@ open class Camera : DefaultNode("Camera"), HasRenderable, HasMaterial, HasCustom
      *
      * Returns (worldPos, worldDir)
      */
-    fun screenPointToRay(x: Int, y: Int): Pair<Vector3f, Vector3f> {
+    fun screenPointToRay(x: Int, y: Int, camOrigin : Boolean): Pair<Vector3f, Vector3f> {
```
moreApi, line 232: @skalarproduktraum Is this switch needed? Line 247 seems to have been a simple mixup or are there cases where we do not want to start the ray directly from the camera?

skalarproduktraum: I am confused by this change. The routine is written in the original way in the way it is, because the camera position differs from the point where the ray starts. If you want to go from a screen position (which always resides on the near plane of the camera, which has a finite extent) to a world-space ray, you can't just start from the (zero-dimensional/infinitely small) camera position. Or am I missing something here?

skalarproduktraum:
@PowerOfNames Please also make sure all the CI tests work when submitting a PR, at the moment some tests fail.

skalarproduktraum (changes requested):
Please make sure the tests work and check if the camOrigin addition to screenPointToRay actually make sense.

src/main/kotlin/graphics/scenery/Camera.kt
PowerOfNames: When not starting at the zero position of the camera and using the near plane, you can't hit pbjects that are between the near plane and the camera, which happens if ouy move closer to the e.g. plane, that displays a menu in our case. 
Also, for the purpose of scene-detection ray casting, why not use the camera position instead of the near plane?

skalarproduktraum:
Superseded by #498

PR #488, opened 2022-09-26
Image.kt changed bufferImageToRGBABuffer imageBuffer allocation to us…

PowerOfNames:
Inside Image.kt

Method: bufferedImageToRGBABuffer ->
Line 144: changed imageBuffer allocation to use MemoryUtil.memAlloc() instead of ByteBuffer.allocateDirect() -> fix memory overflow when repeatetly using the function


skalarproduktraum:
Thank you, @PowerOfNames 👍

PR #497, opened 2022-11-02
Volume: Let currentTimepoint return the correct value

skalarproduktraum:
This PR fixes an issue where `Volume.currentTimepoint` always returned 0. It now returns the correct value, and sets the BVV state variable `numTimepoints` correctly as well.

cc @moreApi

### DIFF
--- a/src/main/kotlin/graphics/scenery/volumes/Volume.kt
+++ b/src/main/kotlin/graphics/scenery/volumes/Volume.kt
@@ -183,7 +183,6 @@ open class Volume(
             }
         }
         set(value) {
-            viewerState.numTimepoints = timepointCount
             viewerState.currentTimepoint = value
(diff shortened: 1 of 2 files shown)


src/main/kotlin/graphics/scenery/volumes/Volume.kt
```diff
@@ -183,8 +183,10 @@ open class Volume(
             }
         }
         set(value) {
+            viewerState.numTimepoints = timepointCount
```
moreApi, line 186: Shouldn't that happen at the point where the amount of timepoints is manipulated?

skalarproduktraum: Good point, I shall change that 👍

moreApi (approved):

PR #498, opened 2022-11-07
Ray casting fix

PowerOfNames:
Accidentaly created a new branch from the previous PR RaycastingFix (https://github.com/scenerygraphics/scenery/pull/485)
I rewinded the changes there and updated the ray origin to use -nearClippingDistance instead of -1.0 as z coord.


skalarproduktraum:
Thanks, @PowerOfNames! Could you please have a look at the tests `testCastRayFromScreenPos` and `testPickFromScreenPos`, both fail - which causes the CI to go red.

skalarproduktraum (approved):

PR #502, opened 2022-11-10
Gitlab CI: Stabilise Argos results by using scenery.RandomSeed property

skalarproduktraum:
This PR should stabilise Argos comparison results, by using the `scenery.RandomSeed` system property for seeding the PRNG.


moreApi (approved):
looks good

PR #503, opened 2022-11-14
Fixed covers for curves and computed normals within the curve class

Domino2357:
The cover calculation for the curve class was erroneous, leaving the empty tubes open for the user to see. The error was tracked down to a confusion of `last()` with `first()` and will be fixed by this pull request.
Moreover, the calculation of the Curves was delegated to `recalculateNormals()`. From this pull request onwards, the calculation will happen in the `Curve` class itself.

### DIFF
--- a/src/main/kotlin/graphics/scenery/geometry/Curve.kt
+++ b/src/main/kotlin/graphics/scenery/geometry/Curve.kt
@@ -255,7 +255,7 @@ class Curve(spline: Spline, partitionAlongControlpoints: Boolean = true, private
                 return Pair(verticesVectors, normalVectors)
             }
             if (addCoverOrTop == 0) {
-                val newVerticesAndNormals = getCoverVertices(curveGeometry.last(), false)
+                val newVerticesAndNormals = getCoverVertices(curveGeometry.first(), true)
                 verticesVectors.addAll(newVerticesAndNormals.first)
                 normalVectors.addAll(newVerticesAndNormals.second)
             }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/graphics/scenery/geometry/Curve.kt
```diff
@@ -242,91 +244,150 @@ class Curve(spline: Spline, partitionAlongControlpoints: Boolean = true, private
...
         /**
-         * This function calculates the triangles for the the rendering. It takes as a parameter
+         * This function calculates the triangles for the rendering. It takes as a parameter
          * the [curveGeometry] List which contains all the baseShapes transformed and translated
          * along the curve.
          */
-        fun calculateTriangles(curveGeometry: List<List<Vector3f>>, addCoverOrTop: Int = 2): ArrayList<Vector3f> {
+        fun calculateTriangles(curveGeometry: List<List<Vector3f>>, addCoverOrTop: Int = 2): Pair<ArrayList<Vector3f>, ArrayList<Vector3f>> {
             val verticesVectors = ArrayList<Vector3f>(curveGeometry.flatten().size * 6 + curveGeometry[0].size + 1)
+            val normalVectors = ArrayList<Vector3f>(verticesVectors.size/3)
```
skalarproduktraum, line 253: This is unfortunately wrong - there has to be one normal vector for each vertex.

skalarproduktraum (changes requested):

src/main/kotlin/graphics/scenery/geometry/Curve.kt
Domino2357: @skalarproduktraum  Check, I did not know how to compute the normals properly.
A valid approach seems to be to use the normals of the triangles surrounding a vertex, add them together, and normalize the final results. First question: is that correct?
If yes, then there is definitely some bug still somewhere because we went from zebra to a full-blown rip look, see below
CurveUniformBSplineExample - 2022-11-15 22 58 27

skalarproduktraum (approved):

skalarproduktraum (approved):

PR #506, opened 2022-11-28
Normal calculation for curve covers

Domino2357:
Contains the normal calculation for the curve covers as well as size allocations for the ArrayList in the general normal calculation for the curve.


skalarproduktraum (approved):
Looks good, thanks @Domino2357 👍

PR #511, opened 2023-01-09
Settings upgrade

PowerOfNames:
This PR makes some changes to the Settings.kt class, adding save and load methods to work with .properties files.
In addition to that, adds the SettingsEditor and SettingsEditorExample files.

The SettingsEditor is a Swing based UI, that presents a loaded setting in a 2 column table, showing the settings name plus its value.
The UI can be used to add new settings, save and load the setting into a .properties file, and refresh the shown settings (to present value changed that could happen during runtime)

The SettingEditorExample is an adaptation of the TexturedCubeExample and shows how to set and reset settings values, aand how to add an updateRoutine to a specific setting (which is called when the setting value changes).

Disclaimer: Same as for the TransferFunctionEditor, the Swing window might be tricky to test.


moreApi (approved):
Looks good to me. Now its time for @skalarproduktraum

src/main/kotlin/graphics/scenery/SettingsEditor.kt
```diff
@@ -0,0 +1,226 @@
...
+import javax.swing.table.DefaultTableModel
+
+/**
+ * @author Konrad Michel <konrad.michel@mailbox.tu-dresden.de>
+ *
+ * A Swing UI, that allows the user to load setting in order to manipulate and add settings.
+ * Allows saving and loading of settings into a .properties file
+ * Allows refreshing of the settings table in order to show changes to settings values that happened during runtime.
+ */
+class SettingsEditor(var settings : Settings, private val mainFrame : JFrame = JFrame("SettingsEditor"), width : Int = 480, height : Int = 500) {
```
moreApi, line 22: please put "@JvmOverloads" in line 21 so it is more convinient to use in Java

moreApi (changes requested):

src/main/kotlin/graphics/scenery/Settings.kt
```diff
@@ -9,30 +13,93 @@ import java.util.concurrent.ConcurrentHashMap
  *
  * @author Ulrik Günther <hello@ulrik.is>
  */
-class Settings(override var hub: Hub? = null) : Hubable {
+class Settings(override var hub: Hub? = null, prefix : String = "scenery.", propertiesFilePath : String? = null) : Hubable {
```
skalarproduktraum, line 16: Might be worth making `prefix` a val so it could be queried later

src/main/kotlin/graphics/scenery/Settings.kt
```diff
@@ -9,30 +13,93 @@ import java.util.concurrent.ConcurrentHashMap
...
     private var settingsStore = ConcurrentHashMap<String, Any>()
     private val logger by LazyLogger()
 
+    var settingsUpdateRoutines : HashMap<String, ArrayList<() -> Unit>> = HashMap()
```
skalarproduktraum, line 20: This should probably be private so it can't be fiddled with externally.

src/main/kotlin/graphics/scenery/Settings.kt
```diff
@@ -9,30 +13,93 @@ import java.util.concurrent.ConcurrentHashMap
...
+    lateinit var propertiesFile : File
```
skalarproduktraum, line 21: I think it might make sense not to give a path as `String` to the class, but instead an InputStream. This way, the properties could be read from whereever, JAR file, URL, etc. That would also remove the necessity for the lateinit var here.

src/main/kotlin/graphics/scenery/SettingsEditor.kt
skalarproduktraum: If confused on how to add this annotation here, see `SceneryBase` for an example 👍

skalarproduktraum (changes requested):
Looks great, I just had a few little things to complain about. Address them and we're good to go! Thanks @PowerOfNames :+1:

src/main/kotlin/graphics/scenery/Settings.kt
```diff
@@ -9,30 +13,93 @@ import java.util.concurrent.ConcurrentHashMap
...
+        else
+            out = FileOutputStream(propertiesFile)
+
+        props.store(out, null)
+    }
+
+    /**
+     * Parses the type from the incoming string, returns the casted value
+     */
+    fun parseType(value:String): Any = when {
```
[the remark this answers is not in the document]
PowerOfNames: I cant be private because I use it in other places too. But I ned to tag it that its used somewhere else (In Microscenery, I need to to parcing there) An alternative would be to move it into a Utilities class, but I think thats not your way to do things with scenery :D

src/main/kotlin/graphics/scenery/Settings.kt
PowerOfNames: I c. I wasnt aware that a InpoutStream is a hashable object

src/main/kotlin/graphics/scenery/SettingsEditor.kt
PowerOfNames: 'This annotation is not applicable to target 'class'' -> Also did not do that in the TransferFunctionEditor?!

new.properties
```diff
@@ -0,0 +1,4 @@
+#Wed Feb 01 16:44:23 CET 2023
```
skalarproduktraum, line 1: Is this file needed?

newnew.properties
```diff
@@ -0,0 +1,5 @@
+#Wed Feb 01 16:33:04 CET 2023
```
skalarproduktraum, line 1: Is this file needed?

skalarproduktraum (changes requested):
Looks good, please have a look at the properties files though. They should either be not there, or in a different place (resources folder, if needed for a test, etc).

newnew.properties
PowerOfNames: whoops, they are not supposed to be there

skalarproduktraum (approved):

PR #513, opened 2023-01-18
Volume: Parallel file reading

skalarproduktraum:
This PR enables parallel file reading in Volume. In the fromPath() routine, parallel reading of scifio's Plane objects is enabled  using Kotlin coroutines. For each worker thread, one ReaderFilter is created. For NIFTi files, e.g, the reading time is cut down to 20-25% of the original time.

The parallel reading behaviour can be disabled by setting the scenery setting `Volume.ParallelReads` to `false`, either by defining that setting programmatically, or by setting a system property, e.g. as command line option, `-Dscenery.Volume.ParallelReads=false`.

### DIFF
--- a/src/main/kotlin/graphics/scenery/volumes/Volume.kt
+++ b/src/main/kotlin/graphics/scenery/volumes/Volume.kt
@@ -34,8 +34,10 @@ import graphics.scenery.numerics.OpenSimplexNoise
 import graphics.scenery.numerics.Random
 import graphics.scenery.utils.LazyLogger
 import graphics.scenery.utils.extensions.times
+import graphics.scenery.utils.forEachIndexedAsync
 import graphics.scenery.volumes.Volume.VolumeDataSource.SpimDataMinimalSource
 import io.scif.SCIFIO
+import io.scif.filters.ReaderFilter
 import io.scif.util.FormatTools
 import mpicbg.spim.data.generic.sequence.AbstractSequenceDescription
 import mpicbg.spim.data.sequence.FinalVoxelDimensions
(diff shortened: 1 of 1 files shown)


src/main/kotlin/graphics/scenery/volumes/Volume.kt
```diff
@@ -760,22 +766,38 @@ open class Volume(
...
-//                    imageData.put(swapEndianUnsafe(reader.openPlane(0, plane).bytes))
-//                }
-//            }
+            val planeSize = bytesPerVoxel * dims.x * dims.y
+            if(hub.get<Settings>()?.get("Volume.ParallelReads", true) == true) {
+                val readers = ConcurrentHashMap<Thread, ReaderFilter>()
+                (0 until reader.getPlaneCount(0)).forEachIndexedAsync { index, plane ->
+                    val thread = Thread.currentThread()
+                    val myReader =
+                        readers.getOrPut(thread) { scifio.initializer().initializeReader(FileLocation(file.toFile())) }
```
moreApi, line 776: If we are putting explanation comments on companion objects incase someone does not know enough kotlin. These 5 lines need at least a paragrah. Because I know a bit of kotlin but am unable to understand it.

For each coroutine we take the current thread. That is which thread? A coroutine worker thread? And put or get it in some hashmap. Why get? And why use thread as the key?

src/main/kotlin/graphics/scenery/volumes/Volume.kt
```diff
@@ -760,22 +766,38 @@ open class Volume(
...
+                    val bytes = reader.openPlane(0, plane).bytes
+                    val view = imageData.duplicate().order(ByteOrder.LITTLE_ENDIAN)
+                    view.put(bytes)
+                }
+
+                val duration = (System.nanoTime() - start) / 10e5
+                logger.debug("Reading took $duration ms, no parallel readers.")
+            }
 
-            imageData.flip()
```
moreApi, line 778: where is the pointer without this line? What is the convention where to leave a pointer after one is done with a buffer?

skalarproduktraum: We're only operating on a duplicate, so imageData's position etc. remains at the original values.

moreApi (approved):
Now I understand what is happening completely. I am just wondering whether coroutine worker threads are stable. Otherwise we could be leaking readers. But I think that is unlikely and not to bad.

skalarproduktraum:
Coroutines use thread pools, so for the period of reading a file, the thread pool should be stable enough for that purpose. Afterwards, the readers are deallocated anyway.

PR #519, opened 2023-02-17
Improvement of startup times and memory optimisation

skalarproduktraum:
This PR adds the following optimisations:
* LazyLogger is not using Kotlin reflection anymore for determining class names, significantly improving memory usage
* In SceneryBase, the REPL is not created and shown by default anymore. Access to the REPL can be obtained by pressing Shift-R when not running in headless mode. This saves around 300MiB of memory needed for Jython initialisation.
* VulkanRenderer is not using ClassGraph anymore for discovery of available special-purporse swapchains. The swapchains shipped with scenery are checked by default, custom swapchains can be added via `VulkanRenderer.registerSwapchain()`
* InputHandler is not using ClassGraph anymore for the discovery of available InputHandlers. The InputHandlers shipped with scenery are available by default, custom InputHandlers can be added via `InputHandler.registerInputHandler()`

On my machine, this improves the startup time of `TexturedCubeExample` from 5s to 1.8s, and the memory usage from around 900 MiB to 250MiB.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -16,7 +16,6 @@ plugins {
     scenery.publish
     scenery.sign
     id("com.github.elect86.sciJava") version "0.0.4"
-    id("org.sonarqube") version "3.3"
     jacoco
     id("com.github.johnrengelman.shadow") version "7.1.0"
 }
@@ -69,7 +68,7 @@ dependencies {
                 // Vulkan binaries are only necessary on macOS
                 p.endsWith("vulkan") -> {
                     if(native.contains("macos")) {
-                        println("vulkan: org.lwjgl:lwjgl$p:$lwjglVersion:$native")
+                        // println("vulkan: org.lwjgl:lwjgl$p:$lwjglVersion:$native")
                         runtimeOnly("org.lwjgl:lwjgl$p:$lwjglVersion:$native")
                     }
                 }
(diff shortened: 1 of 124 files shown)


src/test/kotlin/graphics/scenery/tests/examples/basic/TexturedCubeExample.kt
```diff
@@ -52,6 +52,11 @@ class TexturedCubeExample : SceneryBase("TexturedCubeExample") {
                 Thread.sleep(20)
             }
         }
+
+        thread {
+            Thread.sleep(5000)
+            System.gc()
```
moreApi, line 58: leftover from testing?

moreApi (changes requested):

src/test/kotlin/graphics/scenery/tests/examples/basic/TexturedCubeExample.kt
skalarproduktraum: Indeed, that is totally not supposed to be there. Good catch!

build.gradle.kts
```diff
@@ -69,7 +68,7 @@ dependencies {
                 // Vulkan binaries are only necessary on macOS
                 p.endsWith("vulkan") -> {
                     if(native.contains("macos")) {
-                        println("vulkan: org.lwjgl:lwjgl$p:$lwjglVersion:$native")
+                        // println("vulkan: org.lwjgl:lwjgl$p:$lwjglVersion:$native")
```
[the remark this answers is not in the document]
skalarproduktraum: Just saw that Gradle actually supports log levels. I've replaced those `println` now with `logger.info()`, so the messages will only show up if you run Gradle with `--info` 👍

moreApi (approved):
CI is Red

PR #522, opened 2023-04-14
Ribbon List to Buffer substitution

Domino2357:
This pull request removes superfluous instances of `ArrayList` in the `Curve` class for memory improvement purposes.

The memory performance is not yet stellar in this PR. However, here, the `LazyLogger()` is still active which took the lion's share of the unnecessary memory usage.


skalarproduktraum (approved):
Thanks, @Domino2357 👍

Domino2357:
Thanks @skalarproduktraum ! As we already discussed last meeting, the CI was likely failing because this memory optimization is still incomplete and belongs together with #519 and #528 .

PR #523, opened 2023-05-03
Buffer backend rework, SSBO support

skalarproduktraum:


src/main/kotlin/graphics/scenery/backends/vulkan/VulkanNodeHelpers.kt
```diff
@@ -352,7 +379,14 @@ object VulkanNodeHelpers {
...
      * Returns true if the node has been given a custom shader, and false if not.
      */
-    fun initializeCustomShadersForNode(device: VulkanDevice, node: Node, addInitializer: Boolean = true, renderpasses: Map<String, VulkanRenderpass>, lateResizeInitializers: MutableMap<Renderable, () -> Any>, buffers: VulkanRenderer.DefaultBuffers): Boolean {
+    fun initializeCustomShadersForNode(
+        device: VulkanDevice,
+        node: Node,
+        addInitializer: Boolean = true,
+        renderpasses: Map<String, VulkanRenderpass>,
+        lateResizeInitializers: MutableMap<Renderable, () -> Any>,
+        buffers: VulkanRenderer.DefaultBuffers)
```
skalarproduktraum, line 388: What's the reasoning behind handing the buffers over here? They could be determined from the `node`, not?

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanObjectState.kt
```diff
@@ -44,6 +45,9 @@ open class VulkanObjectState : NodeMetadata {
     /** UBOs required by the [graphics.scenery.Node] this metadata object is attached to. */
     var UBOs = LinkedHashMap<String, Pair<Long, VulkanUBO>>()
 
+    //Like this here?
+    var SSBOs = LinkedHashMap<String, Pair<Long, VulkanUBO>>()
```
skalarproduktraum, line 49: The intent here is to have this work the same way as the UBO hash map, right? So you'd basically store the SSBO name, together with the Vulkan handle, and the actual UBO defining the data structure as value.

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
```diff
@@ -709,6 +711,17 @@ open class VulkanRenderer(hub: Hub,
                 usage = VK_BUFFER_USAGE_VERTEX_BUFFER_BIT or VK_BUFFER_USAGE_INDEX_BUFFER_BIT or VK_BUFFER_USAGE_TRANSFER_DST_BIT
             )
 
+            ssboUploadPool = VulkanBufferPool(
+                device,
+                usage = VK_BUFFER_USAGE_STORAGE_BUFFER_BIT or VK_BUFFER_USAGE_TRANSFER_DST_BIT
```
skalarproduktraum, line 716: Just to double-check: the upload pool is intended to receive data from the CPU side, yes? Then `TRANSFER_DST_BIT` is correct. Otherwise the two need to be switched. Also, is there a particular reason to have two pools? There's likely no performance issue with having one that has both `SRC` and `DST`.

src/main/kotlin/graphics/scenery/primitives/SSBOTest.kt
```diff
@@ -0,0 +1,184 @@
+package graphics.scenery.primitives
```
skalarproduktraum, line 1: This is the completely wrong place for an example 😅 And if it's not an actual test (which seems to be the case) then it should not be named `Test` (this might actually lead to issues with auto-discovery of tests, some of my logic there just filters by class name)

PowerOfNames: Ahh yes, at that point I wasn't sure what to do to test it. Its more a help for me to visualize the general data flow through the renderer ':D

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
PowerOfNames: On the GPU side, there is a slight performance boost, as SSBOs are layed out differently. Also, a read only buffer can live in faster device local memory, while a writable buffer needs to be host visible or coherent (not to familiar with the concepts yet)
Thats why I decided to split up and download buffers into 2. You dont always want to write into a buffer and vice verse. Especially for read only buffers it might be good to keep this practice

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanObjectState.kt
PowerOfNames: Yes, i think that was my train of thought. Although Im not sure if we should refactor UBOS and SSBOs a bit. because Vulkan differentiates between ubos and ssbos, I needto extent the VulkanUBO class anyways. But notsure if this is 'trivially' doable while keepeing backwards-comp.

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanNodeHelpers.kt
PowerOfNames: ':D I think I just did the line breaks there cause it was so long even at my screen. I did not add anything there

PowerOfNames: And the VulkanRenderer.DefaultBuffers contain the scene ubos. The node does not know about them, so that should be the reason you included them there

PR #526, opened 2023-05-17
Fix custom descriptor set creation

skalarproduktraum:
This PR fixes an issue where custom descriptor sets for textures were not correctly created, mixing up binding and array indices. It's kind of a mystery to me how this could have ever worked correctly 😵‍💫

Two further small changes:
* the `NodeMetadata` interface was removed because it's not needed
* `ShaderMaterial`'s API got slightly simplified
* `VulkanDevice` now exposes enabled/disabled device features


moreApi (commented):
I am not a suitable reviewer for this PR. I cant say anthing about the vulcan/open gl code. Otherwise it looks good to me.

aryaman-gupta (approved):
The `CustomVolumeManager` and `ComputeShaderExample` work now, so the previous issue is fixed.

PR #529, opened 2023-05-22
Curve restructuring

Domino2357:
This pull request reorders the Curve class. Specifically, it breaks it down into an Interface and an abstract class. This allows for more flexibility. In our case, it allows the user to choose a single base shape or if they want to trade more flexibly for a slightly worse performance.
Also cleaned up the code a bit, for example, put the coverOrTop attribute into an enum.


Domino2357:
@skalarproduktraum the build on macos throws this error:
`Error: ROR] Cannot create plugin: org.scijava.plugins.scripting.javascript.JavaScriptScriptLanguage`
I tried to track it down, so far to no avail. Do you have a direction you could point me to?

ctrueden:
@Domino2357 wrote:
> `Cannot create plugin: org.scijava.plugins.scripting.javascript.JavaScriptScriptLanguage`
> I tried to track it down, so far to no avail. Do you have a direction you could point me to?

This problem happens when running with Java 17, because Java 17 no longer ships the Nashorn JavaScript engine, so the `JavaScriptScriptLanguage` fails to initialize. You can avoid the problem by commenting out this line from `build.gradle.kts`:
```kotlin
    implementation("org.scijava:scripting-javascript")
```
I'm guessing @skalarproduktraum just has that in there because he likes JavaScript in the REPL. But we need to find a replacement for Java 14+, when they removed Nashorn.

src/main/kotlin/graphics/scenery/geometry/curve/BaseShapesFromSingleShape.kt
```diff
@@ -0,0 +1,20 @@
+package graphics.scenery.geometry.curve
+
+import org.joml.Vector3f
+
+/**
+ * Convinience object to create a curve with only a single base shape.
```
skalarproduktraum, line 6: Typo in convenience

src/main/kotlin/graphics/scenery/geometry/curve/FrenetFrameCalculation.kt
```diff
@@ -0,0 +1,116 @@
...
+import org.joml.Vector3f
+import kotlin.math.acos
+
+
+/**
+ * Computes the Frenet Frames along a curve defined by a spline.
+ * @param spline: spline along which the frenet frames are to be calculated
+ * @param [firstPerpendicularVector] vector to which the first frenet tangent shall be perpendicular to.
+ */
+object FrenetFrameCalculation{
```
skalarproduktraum, line 17: I feel this could be in `Curve`, not?

src/main/kotlin/graphics/scenery/geometry/curve/PartialCurve.kt
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Each child of the curve must be, per definition, another Mesh. Therefore, this class turns a List of
+ * vertices into a Mesh.
+ *
+ * @param verticesBuffer: corresponds to the vertices buffer of Mesh
+ * @param normalVectors: corresponds to the normals buffer of Mesh
+ *
+ */
+class PartialCurve(verticesBuffer: FloatBuffer, normalVectors: FloatBuffer) : Mesh("PartialCurve") {
```
skalarproduktraum, line 15: PartialCurve should maybe be renamed PartialCurveMesh? I feel the current name kinda gives the (wrong) impression it's a kind of `Curve`.

skalarproduktraum (changes requested):
Apart from some minor points in the reorganisation, this looks good 👍  Have a look at my comments!

src/main/kotlin/graphics/scenery/geometry/curve/FrenetFrameCalculation.kt
Domino2357: I am not sure about this one. I don't think it should be in `Curve` because `Curve` is even abstracted from frenet frames (e.g., `Helix` is a Curve that requires no frenet frames to be computed). It could be part of `FrenetOrientedCurve` but if someone were to use a different way of computing the frames, the two components would be again fused.  On the other hand, this is unlikely. So if you agree, then lets go with your intuition!

skalarproduktraum (approved):

PR #531, opened 2023-05-26
Line color fix

PowerOfNames:
This PR fixes an issue with Line-Colors not being rendered properly
Fixes Line color, Starting color and End color behaviour of the Line.kt class

Open: LinePair.kt -> How shoud they behave? The BundleExample looks odd, but not sure how it is supposed to look.


ctrueden:
@PowerOfNames @skalarproduktraum I rebased this over the current mainline.

kephale:
@skalarproduktraum what is the plan with this PR? it looks like it could be pushed through.

PR #532, opened 2023-05-26
Small kool showcase examples

elect86:
As titled:

- added _kool_ library dependency
- switched code in `Geometry::recalculateNormals` and `Icosphere` constructor to pointers
- added some pointers and typealias utils
- using _kool_ constructors and methods to allocate buffers

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -22,6 +22,7 @@ repositories {
     mavenCentral()
     maven("https://maven.scijava.org/content/groups/public")
     maven("https://jitpack.io")
+    maven("https://raw.githubusercontent.com/kotlin-graphics/mary/master")
 //    mavenLocal()
 }
 
@@ -133,6 +134,8 @@ dependencies {
 
     implementation("org.jfree:jfreechart:1.5.0")
     implementation("net.imagej:imagej-ops:0.45.5")
+
+    implementation("kotlin.graphics:kool:0.9.79")
 }
 
 val isRelease: Boolean
--- a/src/main/kotlin/graphics/scenery/Icosphere.kt
+++ b/src/main/kotlin/graphics/scenery/Icosphere.kt
@@ -1,10 +1,19 @@
 package graphics.scenery
 
+import graphics.scenery.utils.Face
 import graphics.scenery.utils.extensions.minus
 import graphics.scenery.utils.extensions.plus
 import graphics.scenery.utils.extensions.times
 import graphics.scenery.utils.extensions.xy
+import graphics.scenery.utils.inc
+import graphics.scenery.utils.set
+import kool.FloatBuffer
+import kool.IntBuffer
+import kool.adr
+import kool.toPtr
+import org.joml.Vector2f
 import org.joml.Vector3f
+import org.lwjgl.system.MemoryUtil
 import java.util.*
 import kotlin.math.*
 
@@ -16,15 +25,15 @@ import kotlin.math.*
  * @param[subdivisions] Number of subdivisions of the base icosahedron
  */
 open class Icosphere(val radius: Float, val subdivisions: Int) : Mesh("Icosphere") {
-    fun MutableList<Vector3f>.addVertex(vararg v: Float) {
-        this.add(Vector3f(v))
+    fun MutableList<Vector3f>.addVertex(x: Float, y: Float, z: Float) {
+        this += Vector3f(x, y, z)
     }
 
-    fun MutableList<Triple<Int, Int, Int>>.addFace(i: Int, j: Int, k: Int) {
-        this.add(kotlin.Triple(i, j, k))
+    fun MutableList<Face>.addFace(i: Int, j: Int, k: Int) {
+        this += Triple(i, j, k)
     }
 
-    protected fun createBaseVertices(vertices: MutableList<Vector3f>, indices: MutableList<Triple<Int, Int, Int>>) {
+    protected fun createBaseVertices(vertices: MutableList<Vector3f>, indices: MutableList<Face>) {
         val s = sqrt((5.0f - sqrt(5.0f)) / 10.0f)
         val t = sqrt((5.0f + sqrt(5.0f)) / 10.0f)
 
@@ -73,11 +82,11 @@ open class Icosphere(val radius: Float, val subdivisions: Int) : Mesh("Icosphere
 
     protected fun refineTriangles(recursionLevel: Int,
                                   vertices: MutableList<Vector3f>,
-                                  indices: MutableList<Triple<Int, Int, Int>>): MutableList<Triple<Int, Int, Int>> {
+                                  indices: MutableList<Face>): MutableList<Face> {
         // refine triangles
         var faces = indices
-        (0 until recursionLevel).forEach {
-            val faces2 = ArrayList<Triple<Int, Int, Int>>(indices.size * 3)
+        repeat(recursionLevel) {
+            val faces2 = ArrayList<Face>(indices.size * 3)
 
             for (triangle in faces) {
                 // replace triangle by 4 triangles
@@ -124,7 +133,7 @@ open class Icosphere(val radius: Float, val subdivisions: Int) : Mesh("Icosphere
         val i = this.addVertex(middle)
 
         // store it, return index
-        middlePointIndexCache.put(key, i)
+        middlePointIndexCache[key] = i
         return i
     }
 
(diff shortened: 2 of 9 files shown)


build.gradle.kts
```diff
@@ -22,6 +22,7 @@ repositories {
     mavenCentral()
     maven("https://maven.scijava.org/content/groups/public")
     maven("https://jitpack.io")
+    maven("https://raw.githubusercontent.com/kotlin-graphics/mary/master")
```
ctrueden, line 25: Can we please get artifacts deployed to maven.scijava.org, instead of a custom GitHub-based repository, which [should not be done](https://stackoverflow.com/a/22210816/1207769)?

src/main/kotlin/graphics/scenery/Icosphere.kt
```diff
@@ -73,11 +82,11 @@ open class Icosphere(val radius: Float, val subdivisions: Int) : Mesh("Icosphere
 
     protected fun refineTriangles(recursionLevel: Int,
                                   vertices: MutableList<Vector3f>,
-                                  indices: MutableList<Triple<Int, Int, Int>>): MutableList<Triple<Int, Int, Int>> {
+                                  indices: MutableList<Face>): MutableList<Face> {
```
ctrueden, line 85: Hooray for type aliasing! :trophy:

src/main/kotlin/graphics/scenery/Icosphere.kt
```diff
@@ -124,7 +133,7 @@ open class Icosphere(val radius: Float, val subdivisions: Int) : Mesh("Icosphere
         val i = this.addVertex(middle)
 
         // store it, return index
-        middlePointIndexCache.put(key, i)
+        middlePointIndexCache[key] = i
```
ctrueden, line 136: @elect86 Do you notice these manually, or lean on an IDE cleanup/simplification action? If so, what is it?

src/main/kotlin/graphics/scenery/Icosphere.kt
```diff
@@ -177,14 +189,10 @@ open class Icosphere(val radius: Float, val subdivisions: Int) : Mesh("Icosphere
...
-                uv2.xy().get(texcoords).position(texcoords.position() + 2)
-                uv3.xy().get(texcoords).position(texcoords.position() + 2)
+                pTxc++[0] = uv1.xy()
+                pTxc++[0] = uv2.xy()
+                pTxc++[0] = uv3.xy()
             }
-
-            vertices.flip()
-            normals.flip()
-            texcoords.flip()
```
ctrueden, line 187: Just to be sure: the kool `FloatBuffer` does not require flipping, whereas the one allocated by `BufferUtils` does?

src/main/kotlin/graphics/scenery/attribute/geometry/Geometry.kt
```diff
@@ -41,38 +45,22 @@ interface Geometry : Serializable {
...
-
-            val v2 = Vector3f(vertexBufferView[i], vertexBufferView[i + 1], vertexBufferView[i + 2])
-            i += 3
-
-            val v3 = Vector3f(vertexBufferView[i], vertexBufferView[i + 1], vertexBufferView[i + 2])
-            i += 3
+        for (i in vertices.indices step 3) {
+            val v1 = pVtx++[0]
+            val v2 = pVtx++[0]
+            val v3 = pVtx++[0]
```
ctrueden, line 54: This is super concise, but I find it harder to understand. Maybe a middle ground could be:
```kotlin
fun makeVector(array, i) { return Vector3f(array[i], array[i + 1], array[i + 2]) }

val v1 = makeVector(vertexBufferView, i += 3)
val v2 = makeVector(vertexBufferView, i += 3)
val v3 = makeVector(vertexBufferView, i += 3)
```
or some such? Or even attach a `toVector(i)` method to the `FloatBuffer` class here?

It is fun to use overloaded operators, but the pointer thing with `[0]` is strange.

src/main/kotlin/graphics/scenery/utils/pointers.kt
```diff
@@ -0,0 +1,31 @@
+package graphics.scenery.utils
+
+import kool.Ptr
+import kool.unsafe
```
ctrueden, line 4: :exclamation: unsafe? Like `sun.misc.Unsafe`? Do you discourage its use? Or...?

src/test/kotlin/graphics/scenery/tests/examples/compute/CustomVolumeManagerExample.kt
```diff
@@ -40,7 +41,7 @@ class CustomVolumeManagerExample : SceneryBase("CustomVolumeManagerExample") {
             ))
         volumeManager.customTextures.add("OutputRender")
 
-        val outputBuffer = MemoryUtil.memCalloc(1280*720*4)
+        val outputBuffer = ByteBuffer(1280*720*4)
```
ctrueden, line 44: I love the conciseness of this. But it makes me nervous that the kool buffer types are named `ByteBuffer` and `FloatBuffer`, when `java.nio` also has classes with those names. Does the possibility of name clashes or reader confusion concern you? If so, you could rename the kool buffer types to something else like `KoolByteBuffer` and `KoolFloatBuffer`, or `Uint8Buffer` and `Float32Buffer`.

ctrueden (commented):
Here are some thoughts based on my naive reading of the code, as a Kotlin newbie. Maybe some of it helps, I dunno! :shrug: :+1:

build.gradle.kts
elect86: What shouldn't be done is loading big jars (ie fat jars)

This is not the case, the ones I push weight few KBs

src/main/kotlin/graphics/scenery/Icosphere.kt
elect86: If suggested, "Alt+Enter" opens the action dialog and then you select the suggestion

src/main/kotlin/graphics/scenery/attribute/geometry/Geometry.kt
elect86: it's essentially `*pVtx++`

Another alternative:

```
        for (i in vertices.indices step 3) {
            val v1 = pVtx[i]
            val v2 = pVtx[i + 1]
            val v3 = pVtx[i + 2]
```

src/main/kotlin/graphics/scenery/utils/pointers.kt
elect86: Yep, atm it's the way for low level native operations

The whole Lwjgl ultimately rely on that as well

src/test/kotlin/graphics/scenery/tests/examples/compute/CustomVolumeManagerExample.kt
elect86: It's coherent with the current way to allocate Arrays and Lists

`val ints = IntArray(2)`
`val ints = List(2) { i -> i + 1 }`

build.gradle.kts
ctrueden: It shouldn't be done at all IMHO—now you have to run that repo forever or else old builds will break. And as described in that article, it creates a situation where there's another repository that has to be queried for all the dependencies, slowing down builds. Just release the projects on maven.scijava.org or oss.sonatype.org! Once you have the infrastructure in place for Sonatype, deploying is a button push—the only huge PITA is getting it set up at the beginning.

src/main/kotlin/graphics/scenery/Icosphere.kt
ctrueden: Cool, I just wanted to make sure these lines didn't get dropped by accident. :+1:

src/main/kotlin/graphics/scenery/attribute/geometry/Geometry.kt
ctrueden: Nice, I like that much more!

src/test/kotlin/graphics/scenery/tests/examples/compute/CustomVolumeManagerExample.kt
ctrueden: Nice. As long as they are actually returning `ByteBuffer` and `FloatBuffer` objects, that's awesome.

build.gradle.kts
elect86: > now you have to run that repo forever or else old builds will break. 

True, but you know what I have to do to run it forever? Just avoid deleting it

I might be move everything to maven, though ([here](https://github.com/kotlin-graphics/glm/issues/35))

I'd love to deploy to maven.scijava.org, but I have a dev way to release very often and that wasn't compatible with that, or do I recall it wrong?

ctrueden: Publishing to maven.scijava.org is just a standard deploy over WebDAV, I'm sure the Gradle tooling will support it with your credentials both manually and/or automatically. As long as your tooling doesn't try to deploy the same release artifact more than once, which is forbidden due to release artifacts being immutable.

elect86: How much does it take the publication over there? How long before it's available for consumption?

ctrueden: Artifacts published to maven.scijava.org are available instantly. Artifacts published to oss.sonatype.org (or s01.oss.sonatype.org) are available instantly from those URLs, but typically take 10-30 minutes to get mirrored over to Maven Central.

PR #535, opened 2023-06-10
Remove the shadow jar from `publish` task

elect86:
workaround according to https://github.com/johnrengelman/shadow/issues/651

To be tested..


elect86:
I tested this locally and it works

Shadow jar is excluded

skalarproduktraum:
Addressed otherwise, thanks though @elect86 👍

PR #537, opened 2023-06-11
this allows to add mavenLocal by CLI

elect86:
More https://docs.gradle.org/current/userguide/init_scripts.html

Essentially, run it with `-I ./gradle/mavenLocal.init.gradle.kts` and mavenLocal will be add for the current command

@ctrueden


ctrueden:
As with scenerygraphics/sciview#449, I'm going to close this for now.

PR #541, opened 2023-06-12
Remove the OpenGL renderer

skalarproduktraum:
This PR removes the OpenGL renderer together with remaining dependencies on JOGL and ClearGL.


skalarproduktraum:
Found the problem: For the script interpreter, what was formerly known as `Python` is now known as `Python (Jython)` 👍

ctrueden:
> what was formerly known as Python is now known as Python (Jython)

Ahh yeah, sorry this burned you @skalarproduktraum. The goal was to make way for a "Python (CPython)" language built on PyImageJ. With the Jython language called "Python", it was confusing, so we figured it would be better to be clear in the name...

skalarproduktraum:
You could have called it... `Jython` 🙃

build.gradle.kts
```diff
@@ -104,7 +101,10 @@ dependencies {
     api("sc.fiji:bigdataviewer-vistools:1.0.0-beta-28")
 
     //TODO revert to official BVV
-    api("graphics.scenery:bigvolumeviewer:7698a01")
+    api("graphics.scenery:bigvolumeviewer:7698a01") {
+        exclude("org.jogamp.gluegen", "gluegen-rt")
+        exclude("org.jogamp.jogl", "jogl-all")
```
ctrueden: These should be `gluegen-rt-main` and `jogl-all-main` respectively. I tested it, and as written the deps still come in, but if you exclude the main artifacts then they are successfully suppressed.

build.gradle.kts
```diff
@@ -25,16 +25,13 @@ repositories {
 }
 
 dependencies {
-    implementation(platform("org.scijava:pom-scijava:31.1.0"))
-    annotationProcessor("org.scijava:scijava-common:2.88.1")
+    implementation(platform("org.scijava:pom-scijava:35.1.1"))
+    annotationProcessor("org.scijava:scijava-common:2.94.1")
```
ctrueden, line 29: You could consider defining any hardcoded versions as `val`s, e.g.: 
```kotlin
val pom_scijava_version = "35.1.1"
implementation(platform("org.scijava:pom-scijava:$pom_scijava_version"))
```
Since it's much more common to bump dep versions than it is to change dependency groupId/artifactId.

But I don't have a strong opinion.

src/main/kotlin/graphics/scenery/backends/Renderer.kt
```diff
@@ -12,6 +10,7 @@ import kotlinx.coroutines.*
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.selects.select
 import org.lwjgl.system.Platform
+import java.lang.RuntimeException
```
ctrueden, line 13: This is weird... does kotlin really make you import `java.lang` packages?

src/main/kotlin/graphics/scenery/utils/Image.kt
```diff
@@ -10,6 +9,7 @@ import java.awt.image.*
 import java.io.BufferedInputStream
 import java.io.IOException
 import java.io.InputStream
+import java.lang.RuntimeException
```
ctrueden, line 12: This is weird... does kotlin really make you import `java.lang` packages?

src/main/kotlin/graphics/scenery/utils/TGAReader.kt
```diff
@@ -0,0 +1,630 @@
...
+                            var j = 0
+                            while(j < width) {
+                                val colormapIndex = bytes[offset + width * i + j].toInt() and 0xFF - colormapOrigin
+                                var color = -0x1
+                                if(colormapIndex >= 0) {
+                                    val index = 3 * colormapIndex + 18
+                                    val b = palette[index + 0].toInt() and 0xFF
+                                    val g = palette[index + 1].toInt() and 0xFF
+                                    val r = palette[index + 2].toInt() and 0xFF
+                                    val a = 0xFF
```
ctrueden, line 159: There is a lot of repeated code here. You could factor out e.g. the BGR extraction to a helper method that returns a `Triple`.

src/main/kotlin/graphics/scenery/utils/TGAReader.kt
```diff
@@ -0,0 +1,630 @@
...
+                                    color = r shl rs or (g shl gs) or (b shl bs) or (a shl `as`)
```
ctrueden, line 204: Ooh, `shl`. Whyyyyyy? Why not go all the way with it:
```kotlin
color eq r shl rs or lpn g shl gs rpn or lpn b shl bs rpn or lpn a shl bt as bt rpn
```
Get rid of all those pesky symbols! So much better. :triumph:

src/test/kotlin/graphics/scenery/tests/unit/controls/SwingMouseAndKeyHandlerTests.kt
```diff
@@ -221,12 +221,12 @@ class SwingMouseAndKeyHandlerTests {
 
             val modifiers = listOf(
                 "",
-//                "ctrl",
+                "ctrl",
 //                "ctrl alt",
 //                "ctrl shift",
```
ctrueden, line 226: Just curious: why do you test only one modifier at a time, instead of mod combos like these commented out ones?

ctrueden (approved):
Awesome work, @skalarproduktraum! I'm not fully qualified to review everything, but I did my best, and left some comments on things I noticed. Nothing showstopping except for the BVV jogamp artifact exclusions, which I fixed with ed25e1d7a0b3985ae9287e57f24a7d9ccfd6235f to save us time.

src/main/kotlin/graphics/scenery/utils/TGAReader.kt
ctrueden: Haha, I found [KT-1440](https://youtrack.jetbrains.com/issue/KT-1440) where people argued ad nauseum about this, and @elect86 was among those commenting in favor of the symbols. :+1: Glad to know I'm not the only one who prefers `<<` over `shl`.

elect86: That's the typical absurd example of denying perfectly normal features for the fear of abuse

src/main/kotlin/graphics/scenery/utils/TGAReader.kt
skalarproduktraum: Won't do that, that code is just copied over from ClearGL and I don't care too much about it. We need it though, for the reason of being able to read TGA files.

src/test/kotlin/graphics/scenery/tests/unit/controls/SwingMouseAndKeyHandlerTests.kt
skalarproduktraum: There were still some (I think platform-dependent) issues remaining with those. Once they are figured out, you're right, the other combos should also be tested 👍

src/main/kotlin/graphics/scenery/backends/Renderer.kt
skalarproduktraum: It does, but only because this was not using Kotlin's version of it. It was unnecessary and I've removed it.

src/main/kotlin/graphics/scenery/utils/Image.kt
skalarproduktraum: It does, but only because this was not using Kotlin's version of it. It was unnecessary and I've removed it.

build.gradle.kts
skalarproduktraum: I'd rather have it together with the GAV for consistency 👍

PR #542, opened 2023-06-12
adds insideNormals for Sphere

smlpt:
Creates flipped normals when `insideNormals = true` is passed to the constructor, the same way it works as with the Box class.


skalarproduktraum (approved):
Very useful, thank you @smlpt 👍

PR #547, opened 2023-06-13
CI: build only mainline and PRs

ctrueden:
Do not build all topic branches.


skalarproduktraum (approved):
Thank you, that is a very useful change 👍

PR #548, opened 2023-06-14
Feature/kotlinesque

elect86:
As titled, this is an attempt for making Scenery more pleasuring for Kotlin users

I tried to leave the style as much as possible untouched


ctrueden:
Rebased over current mainline.

src/test/kotlin/graphics/scenery/tests/examples/basic/ArcballExample.kt
```diff
@@ -19,60 +19,56 @@ class ArcballExample : SceneryBase("ArcballExample") {
...
 
-        val camlight = PointLight(3.0f)
-        camlight.intensity = 5.0f
-        cam.addChild(camlight)
-
-        val box = Box(Vector3f(1.0f, 1.0f, 1.0f))
+        val camlight = PointLight(3.0f) {
+            intensity = 5.0f
+        }
+        cam += camlight
```
ctrueden, line 36: This is a neat trick. But I find it odd to override `plusAssign` but not `plus`. The implication, which is not true for these types, is that `cam + camlight` would be a new camera object with the light added, and only because the `+=` was used does it get assigned back into the same `cam` variable. Is this a pattern Kotlin users often use for mutators, to override only the `+=` but not `+`?

src/test/kotlin/graphics/scenery/tests/examples/basic/ArcballExample.kt
```diff
@@ -19,60 +19,56 @@ class ArcballExample : SceneryBase("ArcballExample") {
...
+            }
+            material {
+                diffuse = Vector3f(1.0f, 1.0f, 1.0f)
+            }
         }
-        scene.addChild(floor)
+        scene += floor
 
-        lights.forEach(scene::addChild)
+        scene += lights
```
ctrueden, line 71: Assuming the `+=` trick is deemed kosher by @skalarproduktraum, the fact that you can `+=` a whole list of children at once is very cool!

ctrueden (commented):
I have no strong opinions about what is more kotlinesque, so I just made a couple of random comments. :game_die:

src/test/kotlin/graphics/scenery/tests/examples/basic/ArcballExample.kt
elect86: Yep

ctrueden: Huh, I guess it's convenient, so I'll swallow my feelings of mathematical discomfort :laughing:

elect86: We might say this is one of the example on the border line, some people will assume this to be an operator abuse, others don't

But that's where I like to give the choice :)

kephale:
@skalarproduktraum @elect86 can this PR be wrapped up easily?

PR #555, opened 2023-06-15
OpenEXR support

smlpt:
- Adds an OpenEXR image reader to the `Image.fromStream()` method.
- Changes the returned texture of `Texture.fromImage()` to use the actual `image.type` instead of always `UnsignedByteType()` to make it compatible with 32bit OpenEXR images.


src/main/kotlin/graphics/scenery/backends/vulkan/VulkanTexture.kt
```diff
@@ -327,13 +329,24 @@ open class VulkanTexture(val device: VulkanDevice,
                 }
 
                 val storage = memAlloc(data.remaining() / 3 * 4)
-                val view = data.duplicate()
-                val tmp = ByteArray(pixelByteSize * 3)
-                val alpha = (0 until pixelByteSize).map { 255.toByte() }.toByteArray()
+                val view = data.duplicate().order(ByteOrder.LITTLE_ENDIAN)
+                val tmp = ByteArray(channelBytes * 3)
+                val alpha = when(gt.type) {
+                    is UnsignedByteType -> (0 until channelBytes).map { 255.toByte() }.toByteArray()
```
skalarproduktraum, line 335: I know this is my doing, but it'd be great if you could add the correct byte sequence for each type here.

src/main/kotlin/graphics/scenery/utils/Image.kt
```diff
@@ -54,36 +72,134 @@ open class Image(val contents: ByteBuffer, val width: Int, val height: Int, val
...
+                    logger.debug("------------------------------")
+                    logger.debug("Level x: " + image.level_x())
+                    logger.debug("Level y: " + image.level_y())
+                    logger.debug("Images: " + image.images()!!.remaining())
+                    logger.debug("Width: " + image.width())
+                    logger.debug("Height: " + image.height())
+                    logger.debug("Number of channels: " + image.num_channels())
+                    logger.debug("Number of tiles: " + image.num_tiles())
+
+                    //FreeEXRImage(image)
```
skalarproduktraum, line 168: Why is this commented out?

src/main/kotlin/graphics/scenery/utils/Image.kt
```diff
@@ -54,36 +72,134 @@ open class Image(val contents: ByteBuffer, val width: Int, val height: Int, val
...
+                    TinyEXR.FreeEXRHeader(header)
+
+                    val pb = image.images()!!
```
skalarproduktraum, line 171: There should be a check here if `images()` actually has a non-null return value. Scanline images would not have iirc. Please issue a warning or exception in that case.

src/main/kotlin/graphics/scenery/utils/Image.kt
```diff
@@ -54,36 +72,134 @@ open class Image(val contents: ByteBuffer, val width: Int, val height: Int, val
...
+                            val b = MemoryUtil.memByteBuffer(pb[c]+i*4, 4)
+                            tmp.put(0, b.get(3))
+                            tmp.put(1, b.get(2))
+                            tmp.put(2, b.get(1))
+                            tmp.put(3, b.get(0))
+                            bb.putFloat(tmp.asFloatBuffer().get(0))
+                        }
+                        bb.putFloat(1.0f)
+                    }
+                    bb.flip()
```
skalarproduktraum, line 186: `stream.close()` is missing in this branch.

src/main/kotlin/graphics/scenery/utils/Image.kt
```diff
@@ -54,36 +72,134 @@ open class Image(val contents: ByteBuffer, val width: Int, val height: Int, val
...
-                    val height = TGAReader.getHeight(buffer)
-                    bi = BufferedImage(width, height, BufferedImage.TYPE_INT_ARGB)
-                    bi.setRGB(0, 0, width, height, pixels, 0, width)
-                } catch (e: IOException) {
-                    Colormap.logger.error("Could not read image from TGA. ${e.message}")
-                    bi = BufferedImage(1, 1, BufferedImage.TYPE_INT_ARGB)
-                    bi.setRGB(0, 0, 1, 1, intArrayOf(255, 0, 0), 0, 1)
+            when {
+
+                extension.lowercase().endsWith("tga") -> {
```
skalarproduktraum, line 77: The `extension.lowercase()` could be moved directly to the `when` as it's the same for each branch.

skalarproduktraum (changes requested):
Looking mostly good, big thanks for a cool addition @smlpt 👍 Please address the changes, then we're good to go 😎

src/main/kotlin/graphics/scenery/utils/Image.kt
```diff
@@ -54,36 +72,134 @@ open class Image(val contents: ByteBuffer, val width: Int, val height: Int, val
...
-                    bi.setRGB(0, 0, 1, 1, intArrayOf(255, 0, 0), 0, 1)
+                extension.lowercase().endsWith("exr") -> {
+                    val memory: ByteBuffer
+                    try {
+                        val array = stream.readAllBytes()
+                        memory = MemoryUtil.memAlloc(array.size)
+                        memory.put(array)
+                        memory.flip()
+                    } catch (e: IOException) {
+                        throw RuntimeException(e)
```
skalarproduktraum, line 105: Don't just re-throw exceptions, that is not useful.

skalarproduktraum (approved):

PR #557, opened 2023-06-16
TransferFunctionEditor: change from JFrame to JPanel

moreApi:


skalarproduktraum (approved):
Thanks @moreApi, this is very useful!

PR #560, opened 2023-06-16
Fixing the visual kinks in the curve geometry

Domino2357:
**This pull request fixes the visual kinks sometimes shown when rendering curves.** 

_What was the problem?_
Splines are defined to be continuous (even differentiable) at the transitions between sections. So, one can define (mathematically) the parameter t piecewise in open intervals (0 <= t <= 1, 1 <= t <= 2, ...). However, in the implementation, we need to make sure that the transition points won't be added twice. We did not do this so far, leading to the computing of two identical base shapes overlapping. 

_What was the solution?_
In principle very simple, compute the splines s.t. the transition points belong to either of two consecutive sub-splines.

_What has been done concretely?_
- complete refactoring of the BSpline computation. We are using matrix multiplication now instead of writing out equations, leading to more readable code.
- adding an extra flag to the sub-spline computation, indicating whether to include the first point or not
- adapting the `Curve` class according to changes in the `UniformBspline` class
- adapting the `RibbonDiagram` to new changes
- cleaning up the `RibbonDiagram`, especially the computation of the sub-shapes, leading to better readability
- adding unit test for the curve class
- adapting the `RibbonDiagramTests` to the changes


Domino2357:
@skalarproduktraum this is ready for review :)

src/main/kotlin/graphics/scenery/proteins/RibbonDiagram.kt
```diff
@@ -522,25 +521,60 @@ class RibbonDiagram(val protein: Protein, private val displaySS: Boolean = false
...
-                    val nextGuide = guidePointList[index + 1]
-                    //Secondary structures which are not sheets or helices are summarized
-                    if (guide.type == SecStrucType.helix4 && nextGuide.type == SecStrucType.helix4 ||
-                            guide.type.isBetaStrand && nextGuide.type.isBetaStrand ||
-                            (guide.type != SecStrucType.helix4 && nextGuide.type != SecStrucType.helix4
-                                    && !guide.type.isBetaStrand && !nextGuide.type.isBetaStrand)){
-                        count++
-                    } else {
-                        return count
+    enum class SimplifiedType { sheet, helix, coil }
```
skalarproduktraum, line 528: Members of enums should be capitalised (e.g. `Sheet`, `Helix`, etc.)

src/main/kotlin/graphics/scenery/proteins/RibbonDiagram.kt
```diff
@@ -522,25 +521,60 @@ class RibbonDiagram(val protein: Protein, private val displaySS: Boolean = false
...
+
+    /**
+     * Stores a section of a secondary structure
+     */
+    data class SSsection(val type: SecStrucType, val length: Int, val subSpline: List<Vector3f>)
```
skalarproduktraum, line 533: How about `SecondaryStructureSection` instead?

src/test/kotlin/graphics/scenery/tests/unit/RibbonDiagramTests.kt
```diff
@@ -169,6 +172,19 @@ class RibbonDiagramTests {
         assertEquals(residuesCount, allResidues.size)
     }
 
+    /**
+     * Checks that the important parts of proteins remain unchanged during ribbon diagram calculation
```
skalarproduktraum, line 176: Why should they change?

src/main/kotlin/graphics/scenery/proteins/RibbonDiagram.kt
skalarproduktraum, line 1: Could you please rename `displaySS` to something more descriptive? `showSecondaryStructure`, for example?

src/main/kotlin/graphics/scenery/proteins/RibbonDiagram.kt
```diff
@@ -118,7 +118,9 @@ class RibbonDiagram(val protein: Protein, private val displaySS: Boolean = false
 
         val subParent = Mesh("SubProtein")
 
-        val splinePoints = spline.splinePoints().map{ it.sub(centroid) }
+        //the very first point of the spline is neglected as this makes the size of the point list divisible by the
+        //section vertices count
+        val splinePoints = spline.splinePoints().map{ it.sub(centroid) }.drop(1)
```
skalarproduktraum, line 123: Dropping one element might be a huge performance hit. Use e.g. a subList of splinePoints() instead

skalarproduktraum (changes requested):
Thanks, only requesting some slight changes :)

src/test/kotlin/graphics/scenery/tests/unit/RibbonDiagramTests.kt
Domino2357: This was something I checked during debugging the `RibbonDiagram`, the chain of amino acids in the BioJava class is mutable. Since we use this list in our code in multiple ways, I figured it'd be a good idea to verify this.

src/main/kotlin/graphics/scenery/proteins/RibbonDiagram.kt
Domino2357: Check. The enum became superfluous anyway, but thanks for the hint!

Domino2357:
@skalarproduktraum this looks good to me. If you find my answer in the open conversation above satisfying, it should be done.

skalarproduktraum (approved):

skalarproduktraum (approved):
Thanks @Domino2357 👍

PR #562, opened 2023-06-19
TransferFunctionEditor -Mig-Layout change to support dynamic resizes + Major BugFix

PowerOfNames:
This PR changes the Mig-Layout of the TransferFunctionEditor to support dynamic window resizes down to 0,0

There was also a major bug, where adding a control point, when the window size was smaller then the internal minimum chart draw size, results in a control point not at mouse-position. This bug was fixed by setting the minimum draw-sizes of the ChartPanel to 0.

Also: Refactored the panel adding to the SwingBridgeFrame, which now packs and sets visibility as well to simplify the API


skalarproduktraum (approved):
Thank you, @PowerOfNames :+1:

And now, that I've also added some changes...

PR #564, opened 2023-06-20
Cylinder: Add top and bottom lids, make smooth shading optional

### LINKED ISSUE #556: Put a lid on cylinders
The cylinders generated by the [Cylinder.kt] class dont have a lid on both ends. They are more of an roll.

smlpt:
Fixes #556 

Refactors the code to work with geometryType Triangles instead of Triangle_Strip to support fan-like face structure of top and bottom caps, default is off (can be turned on with `fillCaps=true`). Normal interpolation can be turned on by passing the attribute `smoothSides`, default is false.

Flat shading with filled caps:
Screenshot 2023-06-20 100418

Smooth shading & texture mapping:
Screenshot 2023-06-20 104840

### DIFF
--- a/src/main/kotlin/graphics/scenery/primitives/Cylinder.kt
+++ b/src/main/kotlin/graphics/scenery/primitives/Cylinder.kt
@@ -3,12 +3,12 @@ package graphics.scenery.primitives
 import graphics.scenery.BufferUtils
 import graphics.scenery.Mesh
 import graphics.scenery.geometry.GeometryType
+import org.joml.Vector2f
 import org.joml.Vector3f
-import java.util.*
+import kotlin.collections.ArrayList
 import kotlin.math.PI
 import kotlin.math.cos
 import kotlin.math.sin
-import kotlin.math.sqrt
 
 /**
  * Constructs a cylinder with the given [radius] and number of [segments].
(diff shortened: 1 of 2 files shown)


src/main/kotlin/graphics/scenery/primitives/Cylinder.kt
```diff
@@ -18,63 +18,95 @@ import kotlin.math.sqrt
  * @param[segments] Number of segments in latitude and longitude.
  */
 
-class Cylinder(var radius: Float, var height: Float, var segments: Int) : Mesh("cylinder") {
+class Cylinder(var radius: Float, var height: Float, var segments: Int, fillCaps: Boolean = false, smoothSides: Boolean = false) : Mesh("cylinder") {
```
skalarproduktraum, line 21: Please add the `@JvmOverloads` annotation to the constructor (it'll look a bit different afterwards, search for the annotation in the scenery code base) - otherwise Java callees will have to always use the full signature, as Java doesn't do default args.

src/main/kotlin/graphics/scenery/primitives/Cylinder.kt
```diff
@@ -18,63 +18,95 @@ import kotlin.math.sqrt
...
     init {
         geometry {
-            geometryType = GeometryType.TRIANGLE_STRIP
+            geometryType = GeometryType.TRIANGLES
```
skalarproduktraum, line 24: Triangle strips are more efficient to render (cache efficiency and so on). The stripification could be done as a postprocessing step, so consider this only a note ^^

src/test/kotlin/graphics/scenery/tests/unit/VolumeMeasuererTests.kt
```diff
@@ -30,6 +30,6 @@ class VolumeMeasurerTests {
     fun testVolumeCylinder() {
         val c = Cylinder(2f, 10f, 5)
         val volume = VolumeMeasurer().calculateVolume(c)
-        assertEquals(volume, 63.40377712249756)
+        assertEquals(volume, 63.40377712249756, absoluteTolerance = 0.00001)
```
skalarproduktraum, line 33: Cut the expected number short here as well please, I'd say something like `63.403` is totally sufficient here.

skalarproduktraum (changes requested):
Some tiny remarks, thanks for the useful contribution @smlpt 👍

skalarproduktraum (approved):
Thanks, @smlpt 👍

PR #565, opened 2023-06-20
CI: Adjust pagefile size on Windows to avoid unnecessary noise

skalarproduktraum:
This PR adjusts the pagefile size on the Github Actions Windows runners, to avoid random out-of-memory errors for memory-intensive unit tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Build scenery
-        run: ./gradlew build -x dokkaHtml -x dokkaHtmlJar -x javadoc -x dokkaJavadocJar --info --no-daemon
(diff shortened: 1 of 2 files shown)


elect86:
What kind of noise are you referring to?

elect86 (approved):
Fine for me

skalarproduktraum:
Some builds on Windows were randomly running out of memory, especially when running protein tests, which are quite memory-consuming, but optimised to a reasonable extent now. I've now searched **_THE INTERNET_** if people experience similar issues, and yes, they did. Problem is solved by increasing the pagefile size, which this PR does 👍 

I'm also limiting info logging more and have optimised the output options for the Gradle test runner, so exceptions are much easier to spot in the logs 🚀

skalarproduktraum:
Thanks for the review, @elect86 👍

elect86:
This was the issue with the "Window 17" platform?

PR #576, opened 2023-07-05
GLFWMouseAndKeyHandler.kt: Ignore modifier key presses themselves. Th…

### LINKED ISSUE #566: Shift key does not register in input handler
GLFW does register shift key events, but they won't reach the behavior. E.g. fast movement with Shift + WASD is not possible; the movement simply stops after the first small step.

Maybe this is related to the changes @skalarproduktraum made last week in commit 845224e?

moreApi:
GLFWMouseAndKeyHandler.kt: Ignore modifier key presses themselves. They are already included in events of the other keys

fixes #566

### DIFF
--- a/src/main/kotlin/graphics/scenery/controls/GLFWMouseAndKeyHandler.kt
+++ b/src/main/kotlin/graphics/scenery/controls/GLFWMouseAndKeyHandler.kt
@@ -64,8 +64,28 @@ open class GLFWMouseAndKeyHandler(var hub: Hub?) : MouseAndKeyHandlerBase(), Aut
         return mask
     }
 
+
+    /**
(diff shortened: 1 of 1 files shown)


moreApi:
true, but was that possible before?

moreApi:
rebased to main @skalarproduktraum

smlpt:
Works again on my side, thanks for the fix!
Just to clarify, Shift + J/K was never supposed to be a default keybind, right? Because that doesn't work.

smlpt (approved):

moreApi:
> Works again on my side, thanks for the fix! Just to clarify, Shift + J/K was never supposed to be a default keybind, right? Because that doesn't work.

Good Question, maybe @skalarproduktraum knows more.

PR #602, opened 2023-08-21
CI: Gitlab CI optimisation

skalarproduktraum:
This PR changes the behaviour of Gitlab CI:
* runs Gitlab CI jobs only on main and PRs
* skips Argos if job was not successful


skalarproduktraum:
This is broken, therefore closing.

PR #605, opened 2023-08-22
Shader compilation improvements

skalarproduktraum:
* improves command line behaviour of the integrated shader compiler (invoke e.g. via `./gradlew CompileShader --args="path/to/shader.vert"`)
* makes vertex shaders more efficient by using the left eye as default when rendering, removing two additional matrix calculations
* updates compiled SPV versions of the included shaders


buildSrc/src/main/kotlin/scenery/base.gradle.kts
```diff
@@ -102,7 +102,7 @@ tasks {
     register<JavaExec>("compileShader") {
         group = "tools"
         mainClass.set("graphics.scenery.backends.ShaderCompiler")
-        classpath = sourceSets["main"].runtimeClasspath
+        classpath = sourceSets["test"].runtimeClasspath
```
moreApi, line 105: I dont understand this. Do we need shaders only when testing?

src/main/kotlin/graphics/scenery/backends/ShaderCompiler.kt
```diff
@@ -187,16 +203,42 @@ class ShaderCompiler: AutoCloseable, Callable<ByteArray> {
...
+            out.name
         } else {
-            out.writeBytes(bytecode)
+            out.toString()
         }
 
-        return bytecode
+        val code = file.readText()
+        if(code.contains(Regex("#ifdef|#ifndef|#else|#endif"))) {
+            println("⏭️  $inputName -> $outputName SKIPPED ${file.name} due to preprocessor directives in code.")
```
moreApi, line 220: what are/were we doing with files with preprocessor directives?

src/main/kotlin/graphics/scenery/backends/ShaderCompiler.kt
moreApi, line 1: why use println instead of logger?

moreApi (changes requested):
Did not review the changes in shaders

buildSrc/src/main/kotlin/scenery/base.gradle.kts
skalarproduktraum: No, what this block of code does is to register a Gradle task with the name `compileShader`. The change from the main to the test classpath only means that some libraries that are usually only in the test classpath will be available (such as picocli or slf4j).

src/main/kotlin/graphics/scenery/backends/ShaderCompiler.kt
skalarproduktraum: That particular function is only used when invoking the compiler from the command line. This means that files with preprocessor directives are just skipped for offline compilation. When invoked from within the renderer, the files are compiled regularly. This is e.g. due to the renderer or volume renderer adding some `#define`s.

src/main/kotlin/graphics/scenery/backends/ShaderCompiler.kt
skalarproduktraum: As this function is only meant to be used for the command line compiler, it seemed more reasonable to use `println` instead of the full slf4j logging infrastructure.

smlpt (approved):

buildSrc/src/main/kotlin/scenery/base.gradle.kts
skalarproduktraum: I actually thought some more about this and found a better solution now. The classpath I've also reverted to `main`, which makes more sense.

src/main/kotlin/graphics/scenery/backends/ShaderCompiler.kt
skalarproduktraum: Fixed, see my comment in `base.gradle.kts`

PR #613, opened 2023-09-14
Add jvm 17 params to readme

moreApi:


skalarproduktraum:
While it's good to document this, it's not a complete solution. This is now being addressed in #619 👍

PR #616, opened 2023-09-19
VulkanDevice: Fix descriptor pool allocation

skalarproduktraum:
This PR fixes a problem in Vulkan descriptor pool allocation, which lead to the problem that no more descriptors could be allocated.

### DIFF
--- a/src/main/kotlin/graphics/scenery/backends/vulkan/VulkanDevice.kt
+++ b/src/main/kotlin/graphics/scenery/backends/vulkan/VulkanDevice.kt
@@ -362,13 +362,27 @@ open class VulkanDevice(
         vkDestroySemaphore(this.vulkanDevice, semaphore, null)
     }
 
-    data class DescriptorPool(val handle: Long, var free: Int = maxTextures + maxUBOs + maxInputAttachments + maxUBOs) {
+    data class DescriptorPool(
+        val handle: Long,
+        var free: ConcurrentHashMap<Int, Int> = ConcurrentHashMap(
+            mutableMapOf(
+                VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER to maxUBOs,
+                VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC to maxUBOs,
+                VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT to maxInputAttachments,
+                VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER to maxTextures,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/graphics/scenery/backends/vulkan/VulkanDevice.kt
```diff
@@ -362,14 +362,28 @@ open class VulkanDevice(
         vkDestroySemaphore(this.vulkanDevice, semaphore, null)
     }
 
-    data class DescriptorPool(val handle: Long, var free: Int = maxTextures + maxUBOs + maxInputAttachments + maxUBOs) {
+    data class DescriptorPool(
+        val handle: Long,
+        var free: ConcurrentHashMap<Int, Int> = ConcurrentHashMap(
```
elect86, line 367: Does it really need to be `Concurrent`?

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanDevice.kt
```diff
@@ -362,14 +362,28 @@ open class VulkanDevice(
...
+                VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER to maxUBOs,
+                VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC to maxUBOs,
+                VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT to maxInputAttachments,
+                VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER to maxTextures,
+                VK_DESCRIPTOR_TYPE_STORAGE_IMAGE to maxTextures)
+        )
+    ) {
 
         companion object {
             val maxTextures = 2048 * 16
```
elect86, line 378: The first three ones might also be `const`

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanDevice.kt
```diff
@@ -362,14 +362,28 @@ open class VulkanDevice(
...
             val maxUBOs = 2048
             val maxInputAttachments = 32
```
elect86, line 380: The three `Int`s above might also be `const`

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanDevice.kt
skalarproduktraum: Yes, as  it might be accessed by different threads, at least for reading. This is to ensure consistency.

PR #619, opened 2023-09-20
JPMS compatibility

skalarproduktraum:
This PR improves compatibility with the JPMS, removing the need for `--add-opens` VM flags when running with Java 17+.
* bigvolumeviewer is bumped to 0.3.2 to remove an unnecessary reflection, which led to an illegal access block (see bigdataviewer/bigvolumeviewer-core#19)
* a `module-info.java` is added, which includes the `jdk.unsupported` package, enabling access to `Unsafe`


build.gradle.kts
```diff
@@ -113,6 +116,8 @@ dependencies {
     api("sc.fiji:bigdataviewer-core:10.4.14")
     api("sc.fiji:bigdataviewer-vistools:1.0.0-beta-28")
     api("sc.fiji:bigvolumeviewer:0.3.3") {
+        exclude("sc.fiji", "bigdataviewer-core")
```
ctrueden, line 119: I would suggest to add a comment explaining what it accomplishes to exclude bigdataviewer-core and bigdataviewer-vistools here, when they are included as direct dependencies immediately above.

build.gradle.kts
```diff
@@ -165,6 +170,10 @@ tasks {
     withType<JavaCompile>().all {
         targetCompatibility = project.properties["jvmTarget"]?.toString() ?: "21"
         sourceCompatibility = project.properties["jvmTarget"]?.toString() ?: "21"
```
ctrueden, line 172: Not a thing with this PR specifically, but maybe rather than setting `-target` and `source`, you want to configure the `--release`?
```kotlin
    options.compilerArgs.add("--release")
    options.compilerArgs.add(project.properties["jvmTarget"]?.toString() ?: "21")
```
Why? [Because](https://stackoverflow.com/q/43102787/1207769).

src/main/java/module-info.java
```diff
@@ -0,0 +1,57 @@
...
+    exports graphics.scenery.net;
+    exports graphics.scenery.numerics;
+    exports graphics.scenery.primitives;
+    exports graphics.scenery.proteins;
+    exports graphics.scenery.repl;
+    exports graphics.scenery.serialization;
+    exports graphics.scenery.textures;
+    exports graphics.scenery.ui;
+    exports graphics.scenery.utils;
+    exports graphics.scenery.volumes;
```
ctrueden, line 20: If I'm reading it right, it looks like this is exporting *all* the `graphics.scenery.*` packages, but none of the `graphics.scenery.*.*` ones? Specifically: missing are `attribute.geometry`, `attribute.material`, `attribute.renderable`, `attribute.spatial`, `backends.software`, `backends.vulkan`, `controls.behaviours`, `controls.eyetracking`, `utils.extensions`. Intentional, or no?

And are you sure you want to expose so much API surface? I guess since scenery hasn't hit 1.0.0 yet, you can still break things. But this work is an opportunity to consider whether it makes sense to create any internal (non-exported) packages, no?

src/test/kotlin/graphics/scenery/tests/examples/volumes/OutOfCoreRAIExample.kt
```diff
@@ -18,10 +18,10 @@ import net.imglib2.img.Img
...
 import net.imglib2.type.numeric.integer.UnsignedShortType
 import net.imglib2.type.volatiles.VolatileUnsignedShortType
-import org.janelia.saalfeldlab.n5.GzipCompression
-import org.janelia.saalfeldlab.n5.N5FSReader
-import org.janelia.saalfeldlab.n5.N5FSWriter
-import org.janelia.saalfeldlab.n5.imglib2.N5Utils
+//import org.janelia.saalfeldlab.n5.GzipCompression
+//import org.janelia.saalfeldlab.n5.N5FSReader
+//import org.janelia.saalfeldlab.n5.N5FSWriter
+//import org.janelia.saalfeldlab.n5.imglib2.N5Utils
```
ctrueden, line 24: So, scenery only uses (used) n5 in tests? Please let me know if you find a good way of dealing with test-scope dependencies with JPMS. We have found it very challenging with SciJava Ops; we were unable to make a separate `module-info.java` in the `src/test` subtree work (see scijava/scijava@ae39206f8fe6eae2aa2da3b0557e0ecc003e1329 for details).

ctrueden (commented):
Various thoughts!

PR #631, opened 2023-10-23
Cam rotation and stats

aryaman-gupta:
Adds small miscellaneous functionalities: 1) use the `ArcballCamera` to rotate camera about a target by a fixed angle, 2) clearing an item recorded by the `Statistics` class. Extends `fromPathRaw` function in `Volume` to transparently handle volume buffers larger than 2 GB

### DIFF
--- a/src/main/kotlin/graphics/scenery/volumes/Volume.kt
+++ b/src/main/kotlin/graphics/scenery/volumes/Volume.kt
@@ -67,13 +67,10 @@ import java.util.concurrent.ConcurrentHashMap
 import java.util.concurrent.CopyOnWriteArrayList
 import java.util.concurrent.atomic.AtomicInteger
 import kotlin.io.path.name
-import kotlin.math.abs
-import kotlin.math.max
-import kotlin.math.min
-import kotlin.math.sqrt
 import kotlin.properties.Delegates
 import kotlin.streams.toList
 import net.imglib2.type.numeric.RealType
+import kotlin.math.*
 
 @Suppress("DEPRECATION")
 open class Volume(
@@ -965,9 +962,10 @@ open class Volume(
                 val id = v.fileName.toString()
                 val buffer: ByteBuffer by lazy {
 
-                    logger.debug("Loading $id from disk")
-                    val buffer = ByteArray(1024 * 1024)
-                    val stream = FileInputStream(v.toFile())
+                logger.debug("Loading $id from disk")
+                val buffer = ByteArray(1024 * 1024)
+                val stream = FileInputStream(v.toFile())
+
                 val numBytes = if(is16bit) {
                     2
                 } else {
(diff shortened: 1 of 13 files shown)


src/main/kotlin/graphics/scenery/volumes/Volume.kt
```diff
@@ -940,7 +986,12 @@ open class Volume(
          *
          * Returns the new volume.
          */
-        @JvmStatic fun fromPathRaw(file: Path, hub: Hub): BufferedVolume {
+        @JvmStatic @JvmOverloads
+        fun <T: RealType<T>> fromPathRaw(
+            file: Path,
+            hub: Hub,
+            type: T
```
skalarproduktraum, line 993: I'd add a default argument here, where `type: T = UnsignedByteType()` 👍 
That'd also make the changes in the example unnecessary.

src/main/kotlin/graphics/scenery/volumes/Volume.kt
```diff
@@ -963,32 +1014,92 @@ open class Volume(
...
+
+            return parent to children
+        }
+
+        /**
+         * Positions [volumes] back-to-back without gaps, using their [pixelToWorld] ratio. Can, e.g., be used
+         * with [fromPathRawSplit] to load volume files greater than 2 GiB into sliced partitions and place
+         * the partitions back-to-back, emulating a single large volume in the scene.
+         */
+        fun positionSlices(volumes: List<Volume>, pixelToWorld: Float) {
```
skalarproduktraum, line 1093: I am not a fan of having this as a static method. How about an extension function of `Volume` or `RichNode`, so you can call this directly on the parent node? Also, `pixelToWorld` should get the default value `volumes.first().pixelToWorldRatio`, so you don't always have to hand it over.

skalarproduktraum (changes requested):
Looks good, please only address the two minor things I've marked 👍

src/main/kotlin/graphics/scenery/volumes/Volume.kt
aryaman-gupta: Sure, moved the `positionSlices` function to `RichNode` where it is called `positionVolumeSlices` and does not take `pixelToWorld` as parameter

src/main/kotlin/graphics/scenery/volumes/Volume.kt
aryaman-gupta: I'm not sure if it is possible to provide a default argument for a template parameter? The code does not compile with the exact change you proposed @skalarproduktraum

skalarproduktraum (approved):

skalarproduktraum (approved):

skalarproduktraum (approved):

PR #632, opened 2023-10-23
CI: split build task into build and test for better log readability

moreApi:
as discussed here https://imagesc.zulipchat.com/#narrow/stream/327495-scenery.2Bsciview/topic/CI.20log.20readability.3A.20split.20build.20and.20test

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,7 +33,9 @@ jobs:
       - name: Grant execute permission for gradlew
         run: chmod +x gradlew
       - name: Build scenery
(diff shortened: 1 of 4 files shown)


skalarproduktraum:
I've made some small changes. The script is now split into a build and test part, as per your suggestion. However, the tasks were not entirely correct, which I've fixed now. Also, I have added a summary list of tests that have failed or were skipped during execution. This looks like the following on Github Actions now:

image

skalarproduktraum (approved):
Thanks @moreApi 👍

PR #642, opened 2023-11-02
Improve shader hot reloading

skalarproduktraum:
This PR improves hot reloading of shaders:

* when any shader file is requested, the shader compiler will look for that shader in the current working directory, and in the `shaders/` subdirectory - the first file found will be watched for changes and reloaded on-the-fly
* hot-reloadable shaders from the filesystem will only be marked as stale if their compilation succeeds
* hot-reloadable shaders will not be cached

Nice hack for Linux and macOS: symlink `src/main/resources/graphics/scenery/backends/shaders` to `./shaders`, so all of scenery's default shaders become hot-reloadable.

### DIFF
--- a/src/main/kotlin/graphics/scenery/backends/Shaders.kt
+++ b/src/main/kotlin/graphics/scenery/backends/Shaders.kt
@@ -1,5 +1,6 @@
 package graphics.scenery.backends
 
+import bvv.core.shadergen.Shader
 import graphics.scenery.utils.lazyLogger
 import kotlinx.coroutines.*
 import java.io.File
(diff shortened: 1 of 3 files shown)


src/main/kotlin/graphics/scenery/backends/Shaders.kt
```diff
@@ -48,7 +49,9 @@ sealed class Shaders() {
      * [shaders], which are assumed to be relative to a class [clazz].
      */
     open class ShadersFromFiles(val shaders: Array<String>,
-                           val clazz: Class<*> = Renderer::class.java) : Shaders() {
+        val clazz: Class<*> = Renderer::class.java) : Shaders() {
```
PowerOfNames, line 52: are those 'zz's intentional?

PR #643, opened 2023-11-02
Improve JInput native loading

skalarproduktraum:
This PR improves loading the native libraries required for JInput. It now uses default Java classpath file access, instead of the obsolete method of trying to find the native JAR from the classpath and extracting it.

In addition, this also should fix the spurious warnings about the JInput native JAR not being found.


PowerOfNames (approved):
Works for me

PR #650, opened 2023-11-12
Implemented Kotlin Assignment Operator

elect86:
plus:
- replaced deprecated `buildDir` with `layout.buildDirectory`
- simplified some constructs for adding jvm args
- using the available `inceptionYear` property instead of a generic entry in properties


skalarproduktraum (approved):
Thanks @elect86 👍

PR #675, opened 2024-02-01
Switch default build to Java 21

skalarproduktraum:
This PR switches the build to Java 21 by default, and will also generate packages with Java 21 bytecode version. In addition, all Gitlab CI builds now use [the same Docker image](https://hub.docker.com/layers/scenerygraphics/nvidia-vulkan/1.3.261.1-ubuntu20.04-v3/images/sha256-ea8c95427e2db6a1623fc989d1251732844a48bbffa276f162cb93ac94080863?context=explore), which includes all prerequisites, such as the JDK, Vulkan SDK, etc. Before, the non-GPU builds used a default Ubuntu image, which required package installations on each run.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,9 +17,7 @@ jobs:
       matrix:
         include:
           - os: ubuntu-latest
-            java: 11
-          - os: ubuntu-latest
-            java: 17
+            java: 21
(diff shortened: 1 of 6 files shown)


moreApi (approved):
LGTM, though I am no build expert. 

I like using a prebuild image that should save some resources. 🍃

PR #676, opened 2024-02-01
Per-node UBO population and per-frame cam state access

skalarproduktraum:
This PR introduces:
* the `PopulatesUBO` interface, which enables node to define their own serialisation procedure, aka, how to translate a node's information into something accessible to shaders that goes beyond ShaderProperties
* access to a given frame's camera state in the Vulkan renderer's render() function.

### DIFF
--- a/src/main/kotlin/graphics/scenery/DefaultNode.kt
+++ b/src/main/kotlin/graphics/scenery/DefaultNode.kt
@@ -48,7 +48,6 @@ open class DefaultNode(name: String = "Node") : Node, Networkable {
         return true
     }
 
-    override var nodeType = "Node"
     override var boundingBox: OrientedBoundingBox? = null
     override val logger by lazyLogger()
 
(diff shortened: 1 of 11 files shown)


src/main/kotlin/graphics/scenery/DefaultNode.kt
```diff
@@ -48,7 +48,6 @@ open class DefaultNode(name: String = "Node") : Node, Networkable {
         return true
     }
 
-    override var nodeType = "Node"
```
smlpt, line 51: Will this cause any trouble in other parts of scenery that rely on this node type?

src/main/kotlin/graphics/scenery/DetachedHeadCamera.kt
```diff
@@ -104,12 +112,36 @@ class DetachedHeadCamera(@Transient var tracker: TrackerInput? = null) : Camera(
...
+                        ?: camSpatial.projection).applyVulkanCoordinateSystem().invert()
+                })
+            }
+            ubo.add("headShift", { hmd?.getHeadToEyeTransform(0) ?: Matrix4f().identity() })
+            ubo.add("IPD", { hmd?.getIPD() ?: 0.05f })
+            ubo.add("stereoEnabled", { cam.stereoEnabled })
+        }
+    }
+
+    var stereoEnabled = false
```
smlpt, line 142: Why is this false it is a stereo camera?

smlpt (changes requested):
Looks good, but maybe add a short documentation to the populate UBO classes?

PR #684, opened 2024-02-09
Don't let volumes defeat push mode

skalarproduktraum:
Simply using `FloatArray.hashCode()` caused the issue that UBO hashing in order to check for updates got defeated by new float arrays with the same contents being allocated. This happened because bigvolumeviewer copies float arrays behind the scenes when assigning them as a uniform. This PR replaces hashCode() calls on FloatArray and IntArray with contentHashCode(), which does not take the specific reference into account, but only the content of the array. 

In addition, this PR also removes repeated allocations of the slicing equation array in Volume, and allocates the array only once.

### DIFF
--- a/src/main/kotlin/graphics/scenery/backends/UBO.kt
+++ b/src/main/kotlin/graphics/scenery/backends/UBO.kt
@@ -351,10 +351,20 @@ open class UBO {
      *
      * Takes into consideration the member's name and _invoked_ value, as well as the
      * buffer's memory address to discern buffer switches the UBO is oblivious to.
+     *
+     * In case the member's _invoked_ value is an IntArray or FloatArray, contentHashCode()
+     * is used instead of hashCode(), such that simple array reallocation without content
+     * changes do not affect the hashing outcome.
      */
     protected fun getMembersHash(buffer: ByteBuffer): Int {
-        return members.map { (it.key.hashCode() xor it.value.invoke().hashCode()).toLong() }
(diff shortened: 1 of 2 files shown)


xulman:
isn't `contentHashCode()` more expensive than the plain `hashCode()`?

xulman:
aha, but it's currently only 64 floats long... = 4 * MAX_SUPPORTED_SLICING_PLANES... that doesn't look too bad
(I was worried that the float array holds the actual volumetric(!) image data...)

moreApi (commented):
looks ok. 

What I dont understand it the purpose of hashCode() if it includes the objects reference. If I care about the objects reference I do a object.equals(other). Because if the references are the same I guarantee you the content is also the same. Is hashCode something that is under our control? Then I would suggest to replace it with that currently is contentHashCode().

skalarproduktraum:
It's quite well-defined how `hashCode()` behaves in Java for arrays (see the Javadoc in java.util.Arrays) - it'll only take into account referential equality. That however means you can change the content, but the reference stays the same, it's the same place in memory, after all.

I think using `contentHashCode()` here is the right way to go, as this is what Kotlin's standard library brings along for exactly this use case.

moreApi (approved):
ok makes sense

skalarproduktraum:
Oh, and hashCode() itself is beyond our control. The one that does the deep comparison (to which `contentHashCode()` is a shortcut) lives in `java.util.Arrays`, while `hashCode()` itself is just the default one that uses referential equality.

PR #691, opened 2024-02-21
Fix uninitialised delegated Nodes

skalarproduktraum:
This PR fixes an issue where Nodes which delegate rendering (such as Volumes) would be left with initialised metadata (from the delegate), but would themselves not be marked initialised. This leads to an issue where they'd cause constant updates and defeat push mode.

To make push mode defeat issues more easily debuggable in the future, this PR also adds `VulkanRenderer.pushModeDefeats`, which keeps track of the reasons for the last five push mode defeats.


smlpt (approved):
looks good

PR #693, opened 2024-02-22
Multiresolution-safe histograms

skalarproduktraum:
This PR adapts the histogram creation code so it's safe for multiresolution volumes.

The logic will now also use a lower-resolution volume to generate histograms, such that large multiresolution volumes are not iterated voxel-by-voxel. By default, a maximum side length of 512 voxels is used. This fixes scenerygraphics/sciview#556.


kephale (approved):
LGTM!

Issue #698, opened 2024-03-05
Resizing transfer function/colormap editor changes rendering of volume


moreApi:
**Describe the bug**
[link to jans nextcloud with example video](https://cloud.viings.de/index.php/s/nYrtYKQCAQCrsDR )

**To Reproduce**
1. have a volume (eg procedual generated volume example)
2. open tf/colormap editor for it
3. resize editor window
4. enjoy color show

**Expected behavior**
Resizing of editor window does not cahnge rendering of volume

moreApi:
Introduced in #688

skalarproduktraum:
This is fixed in #701

Issue #377, opened 2021-02-25
All mouse buttons seem to be recognized as button1


moreApi:
Move the camera works with all mouse buttons but should only be possible with button1. Also, commands wich are assigned to button 2 and 3 do not work.

In my initial investigation I found line 101 in `GLFWMouseAndKeyHandler` to be strange as the pressed button is always set to 0 same goes for the drag event above. However, this might not be relevant as the `getMask` method does not access that value. But that is also the point were I stopped my investigation as my knowledge of AWT input mask is lacking. Maybe the error is there somewhere. I don't see any code adding anything single button 2 and 3 clicks to the mask. And going with the debugger there and pressing left or right mouse buttons results in the same mask (1024).

Also scenerygraphics/sciview#356 of sciviewe might be related.

moreApi:
still valid

kephale:
bump

moreApi:
scrolling also does not work under windows

skalarproduktraum:
I should note here that on macOS, Swing is used instead of GLFW.

PR #701, opened 2024-03-07
Transfer function editor: Fix colormap reassignment and context menu

### LINKED ISSUE #700: Transfer function editor: disable right click placing of new points
**Describe the bug**
Left clicks place new points in the transfer function, and right clicks open a context menu. However, right clicks also place new points, which can mess up the transfer function when you simply want to open the context menu.

**Expected behavior**
Right clicks do not place new points.

skalarproduktraum:
This PR fixes two issues:
* colormaps would be reassigned on redraw events of ColormapPanel. This was unintended and has been changed to reassign only on actual color map changes.
* Right-clicking on the transfer function chart would both open the context menu and draw/modify points. Right-clicking will now only open the context menu.

Fixes #700.


moreApi (approved):
LGTM

is this also related to #698  ?

smlpt (changes requested):
This fixes the right-clicking issues, but now the control points are no longer draggable.

skalarproduktraum:
@moreApi Yes, this fixes #698 as well. I've added that to the description.
@smlpt I've checked - dragging control points still works for me. Are you certain this PR disables dragging for you?

moreApi:
1. nice
2. maybe its related to https://github.com/scenerygraphics/scenery/issues/377 and an OS issue

skalarproduktraum:
Cause identified, see https://stackoverflow.com/questions/17441013/mousedragged-not-returning-appropriate-button-down - bottom line is, `MouseEvent` fired when dragged may or may not contain the button, as there is no state change in the button. Exact behaviour is platform-specific. This is likely unrelated to #377, but I'll check. Thanks for the pointer @moreApi, and thanks for finding this issue, @smlpt 👍

PR #706, opened 2024-03-12
Make atmosphere shader compatible with sciview inspector

smlpt:
This PR adds a toggle and a detach function for keybindings to the Atmosphere class, as well as a flag for the current keybindings state. This allows sciview to enable or disable the sun movement controls. A method to update the emission strength was also introduced, so that the emission strength can be changed from the sciview inspector.


skalarproduktraum:
Closing due to b0rken history. @smlpt will redo the PR based on cherry-picking the relevant commits.

smlpt:
The new PR for this feature is #707

PR #707, opened 2024-03-13
Make atmosphere shader compatible with sciview inspector

smlpt:
This PR adds a toggle and a detach function for keybindings to the Atmosphere class, as well as a flag for the current keybindings state. This allows sciview to enable or disable the sun movement controls. A method to update the emission strength was also introduced, so that the emission strength can be changed from the sciview inspector.


src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
```diff
@@ -17,51 +17,66 @@ import kotlin.math.*
...
- * @param initSunDir [Vector3f] of the sun position. Defaults to sun elevation of the current local time.
- * @param emissionStrength Emission strength of the atmosphere shader. Defaults to 0.3f.
+ * @param initSunDirection [Vector3f] of the sun position. Defaults to sun elevation of the current local time.
+ * @param emissionStrength Emission strength of the atmosphere shader. Defaults to 1f.
  * @param latitude Latitude of the user; needed to calculate the local sun position. Defaults to 50.0, which is central Germany.
  */
-open class Atmosphere(initSunDir: Vector3f? = null, emissionStrength: Float = 0.3f, var latitude: Double = 50.0) :
+open class Atmosphere(
+    initSunDirection: Vector3f? = null,
+    var emissionStrength: Float = 1f,
```
skalarproduktraum, line 26: Write as 1.0f please, for consistency

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
```diff
@@ -17,51 +17,66 @@ import kotlin.math.*
...
         // Spawn a coroutine to update the sun direction
-        val job = CoroutineScope(Dispatchers.Default).launch {
+        job = CoroutineScope(Dispatchers.Default).launch {
             while (!sunDirectionManual) {
-                sunDir = getSunDirFromTime()
-                // Wait 30 seconds
-                delay(30 * 1000)
+                getSunDirectionFromTime()
+                // Wait 10 seconds
+                delay(10 * 1000)
```
skalarproduktraum, line 65: iirc, you can give delay a `Duration`, so you can write `10.seconds`

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
```diff
@@ -17,51 +17,66 @@ import kotlin.math.*
...
+    fun updateEmissionStrength() {
+        material { emissive = Vector4f(0f, 0f, 0f, emissionStrength * 0.3f) }
+    }
+
+    /** Turn the current local time into a sun elevation angle, encoded as cartesian.
      * @param localTime local time parameter, defaults to [LocalDateTime.now].
      */
-    private fun getSunDirFromTime(localTime: LocalDateTime = LocalDateTime.now()): Vector3f {
-        val latitudeRad = toRadians(latitude)
+    fun getSunDirectionFromTime(localTime: LocalDateTime = LocalDateTime.now()) {
```
skalarproduktraum, line 78: The name is deceiving - shouldn't it rather be called `setSunDirectionFromTime`, as it's setting the direction?

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
```diff
@@ -108,18 +123,20 @@ open class Atmosphere(initSunDir: Vector3f? = null, emissionStrength: Float = 0.
...
-            sunDir.rotate(rotation)
+            sunDirection.rotate(rotation)
         }
     }
 
     /** Attach Up, Down, Left, Right key mappings to the inputhandler to rotate the sun in increments.
      * Keybinds are Ctrl + cursor keys for fast movement and Ctrl + Shift + cursor keys for slow movement.
      * @param increment Increment value for the rotation in degrees, defaults to 20°. Slow movement is always 10% of [increment]. */
-    fun attachRotateBehaviours(inputHandler: InputHandler, increment: Float = 20f) {
+    fun attachRotateBehaviors(inputHandler: InputHandler, increment: Float = 20f) {
```
skalarproduktraum, line 133: More nitpicking: I'd just call it `attach/detachBehaviours`, as there are no others.

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
skalarproduktraum: Alternatively, keep it like this, but let it only return the direction, _and_ add a function that calls this one but also sets the direction.

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
skalarproduktraum: emissionStrength should also be a regular member variable, with a custom setter, so you can update the material properties automatically without the need for the `updateEmissionStrength()` method

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
```diff
@@ -17,51 +17,66 @@ import kotlin.math.*
...
     Icosphere(10f, 2, insideNormals = true) {
 
     @ShaderProperty
-    var sunDir: Vector3f
+    var sunDirection = Vector3f(1f, 1f, 1f)
 
+    /** Is set to true if the user manually moved the sun direction. This disables automatic updating.*/
     private var sunDirectionManual: Boolean = false
+    /** Flag that tracks whether the sun position controls are currently attached to the input handler. */
+    var hasControls: Boolean = false
```
skalarproduktraum, line 37: You'll likely not have this one publicly changeable.

skalarproduktraum (changes requested):

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
smlpt: No I think you are right. It used to be a getter, but then I changed the method to directly write into `sunDirection` and forgot to update the method name.

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
```diff
@@ -12,56 +12,73 @@ import java.lang.Math.toRadians
...
 import kotlin.math.*
+import kotlin.time.Duration.Companion.seconds
 
 /**
  * Implementation of a Nishita sky shader, applied to an [Icosphere] that wraps around the scene as a skybox.
  * The shader code is ported from Rye Terrells [repository](https://github.com/wwwtyro/glsl-atmosphere).
  * To move the sun with arrow keybinds, attach the behaviours using the [attachRotateBehaviours] function.
- * @param initSunDir [Vector3f] of the sun position. Defaults to sun elevation of the current local time.
- * @param emissionStrength Emission strength of the atmosphere shader. Defaults to 0.3f.
+ * @param initSunDirection [Vector3f] of the sun position. Defaults to sun elevation of the current local time.
```
skalarproduktraum, line 21: And what happens if you leave it at `null`, as per the default?

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
```diff
@@ -12,56 +12,73 @@ import java.lang.Math.toRadians
...
 
+    /** Is set to true if the user manually moved the sun direction. This disables automatic updating.*/
     private var sunDirectionManual: Boolean = false
+    /** Flag that tracks whether the sun position controls are currently attached to the input handler. */
+    var hasControls: Boolean = false
+        private set
+    // Coroutine job for updating the sun direction
+    private var job = CoroutineScope(Dispatchers.Default).launch(start = CoroutineStart.LAZY) {
+        logger.debug("Launched sun updating job")
+        while (this.coroutineContext.isActive) {
```
skalarproduktraum, line 43: I think this is wrong, you're checking whether the context is active, not the job. It should just be `isActive`, iirc.

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
```diff
@@ -12,56 +12,73 @@ import java.lang.Math.toRadians
...
+            if (!sunDirectionManual) setSunDirectionFromTime()
```
skalarproduktraum, line 44: Full curly brackets here please

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
skalarproduktraum: Also, just fyi, for KDoc, you don't need to use the `@param` annotation from Javadoc, but you can just refer to the parameters as e.g. `[initSunDirection]`.

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
```diff
@@ -12,56 +12,73 @@ import java.lang.Math.toRadians
...
 
         // Spawn a coroutine to update the sun direction
-        val job = CoroutineScope(Dispatchers.Default).launch {
-            while (!sunDirectionManual) {
-                sunDir = getSunDirFromTime()
-                // Wait 30 seconds
-                delay(30 * 1000)
-            }
-        }
+        job.start()
```
skalarproduktraum, line 74: If I understand this correctly, you can't change the behaviour during runtime, right? You need to decide upfront if the sun should move with the time, or be fixed? Could we make this more flexible?

skalarproduktraum (changes requested):
Some details left 👍

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
smlpt: good point, thanks

src/main/kotlin/graphics/scenery/primitives/Atmosphere.kt
smlpt: The job is always launched and it only animates the sun if the flag `sunDirectionManual` is set to false. This was easier than to re-launch the job in a new scope every time. (Maybe it should be renamed to `isSunDirectionManual`?)

The current behavior is this:
- per default, there are no controls attached and the sun moves with time (unless a direction is passed explicitly, which does not happen in the context of sciview)
- the user attaches the controls. This now enables the user to move the sun with the arrow keys. But until they actually moved it, the sun animation continues
- as soon as they moved the sun with the arrow keys, the animation is interrupted
- the flag is set to false again as soon as the controls are detached (= the checkbox is toggled off) and then the animation continues. This will cause the sun to jump back to the position calculated from the current time and latitude.

I can write a summary of this behavior into a comment to make it clear.

skalarproduktraum (approved):
Thank you, @smlpt 👍

PR #717, opened 2024-04-02
Reorganise VDI classes and mark their APIs as `@Experimental`

skalarproduktraum:
This PR reorganises the VDI classes into `volumes.vdi` and marks their API as `@Experimental`. This should not be understood in terms of code quality, but rather so that the API there might still change.


aryaman-gupta:
Looks good to me overall! I also tested the VDI examples and they work, so I'd say this is good to merge!

aryaman-gupta (approved):
Looks good to me overall. I also tested the VDI examples and they work, so I'd say this is good to merge!

PR #730, opened 2024-04-17
Add ConfirmableClickBehaviour

skalarproduktraum:
This adds a ClickBehaviour derivate that needs to be triggered twice to perform an action.


kephale (approved):
LGTM

PR #732, opened 2024-04-17
DeferredLighting: Fix energy conservervation for Oren-Nayar model

skalarproduktraum:
This PR fixes energy conservation when using the Oren-Nayar lighting model.


smlpt:
I can't really give a proper review here without having something to check the formulas against, sorry.

moreApi (changes requested):
pls add a comment to a resource that describes the used algorithm, preferably a url. It should explain what parameters like "CookTorranceVisibility(vec3 N, vec3 V, vec3 H, vec3 L, float roughness) " mean.

PR #733, opened 2024-04-17
Origin/Volume: Introduce support for custom origins and use in Volume

skalarproduktraum:
This PR introduces a new option for setting custom origins of Volumes. Supercedes #731.

### DIFF
--- a/src/main/kotlin/graphics/scenery/volumes/Volume.kt
+++ b/src/main/kotlin/graphics/scenery/volumes/Volume.kt
@@ -166,7 +166,7 @@ open class Volume(
     }
 
     /** What to use as the volume's origin, scenery's default is [Origin.Center], BVV's default is [Origin.FrontBottomLeft]. **/
-    var origin = Origin.Center
+    var origin: Origin = Origin.Center
 
     /** Rendering method */
     var renderingMethod = RenderingMethod.AlphaBlending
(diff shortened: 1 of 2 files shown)


src/main/kotlin/graphics/scenery/volumes/Volume.kt
```diff
@@ -1221,14 +1221,17 @@ open class Volume(
...
             model.scale(volume.localScale())
-            if (volume.origin == Origin.Center) {
-                model.translate(shift)
+            when(val o = volume.origin) {
+                is Origin.Center -> {
+                    val shift = Vector3f(volume.getDimensions()) * (-0.5f)
+                    model.translate(shift)
+                }
+                is Origin.Custom -> model.translate(o.origin)
+                else -> {}
```
smlpt, line 1234: what about FrontBottomLeft? Why is this not taken into account here?

smlpt (changes requested):
looks good, just one question

src/main/kotlin/graphics/scenery/volumes/Volume.kt
skalarproduktraum: The Origins class only has the most common options in there, which is center and FrontBottomLeft. Further adjustments can be made with the Custom choice 👍

PR #737, opened 2024-04-18
Fix assertions in PersistentTextureRequestsExample 

skalarproduktraum:
This PR fixes an issue with semi-deterministic assertions in PersistentTextureRequestsExample and replaces those with deterministic assertions run in a runAfterRender lambda.


src/test/kotlin/graphics/scenery/tests/examples/compute/PersistentTextureRequestsExample.kt
```diff
@@ -108,33 +118,16 @@ class PersistentTextureRequestsExample : SceneryBase("PersistentTextureRequestsE
...
+                    val file = SystemHelpers.dumpToFile(buffer, "texture-${SystemHelpers.formatDateTime(delimiter = "_")}.raw")
+                    assertNotNull(file, "File handle should not be null.")
 
-        thread {
-            while (renderer?.firstImageReady == false) {
-                Thread.sleep(5)
+                    val sum = file.readBytes().sum()
+                    logger.info("Dumped: $file, ${file.length()} bytes, sum=$sum")
+                    assertEquals(file.length(), 1280*720*4, "File should contain the correct number of bytes for a 1280x720xRGBA image")
+                    assert(sum > 3000000) { "Sum of bytes in file should be non-zero" }
```
aryaman-gupta, line 127: Why is the sum compared with 3000000? Is it not sufficient to check that it is not equal to zero? Changes in volume rendering code down the line could change the image generated and therefore the sum

aryaman-gupta (commented):
One minor comment, looks good to me otherwise!

PR #740, opened 2024-04-19
Pass a fixed sun position to the Atmosphere constructor for reproducability

smlpt:
The default sun position was influenced by local time, thus screwing with the Argos tests for the `AtmosphereExample`.
This PR solves this bug by passing a fixed sun position vector to the constructur.


skalarproduktraum:
This is good, but how about adding your keybindings to the example?

smlpt:
> This is good, but how about adding your keybindings to the example?

@skalarproduktraum the keybindings are added in [L96](https://github.com/scenerygraphics/scenery/blob/0d751ce3f02f8a847cd6858a98b8dd0ae4e08300/src/test/kotlin/graphics/scenery/tests/examples/basic/AtmosphereExample.kt#L96)

skalarproduktraum (approved):

skalarproduktraum:
Thanks, @smlpt 👍

PR #741, opened 2024-04-19
SceneryContext: Fix flickering issue when BVV LUTs update

skalarproduktraum:
This PR:
* fixes an issue where the volume rendering would flicker whenever the BVV LUT was updated, which is quite annoying
* this fix double-buffers the LUT and waits for it to become available on the GPU before replacing the old one, fixing the flickering issue


smlpt (approved):
I am not that well versed with the scenery texture pipeline, but from what I understood I didn't see anything wrong.

PR #764, opened 2024-06-21
Fix atmosphere azimuth range and declination formula

smlpt:
Azimuth range was -180-180, which conflicted with the more common range 0-360 that is set as a sciview property for the inspector.
Declination calculation has also been fix and show be more accurate now.


kephale (approved):
LGTM!

PR #775, opened 2024-08-27
Allow to center camera messages

smlpt:
This adds a new `centering` parameter to `Camera.showMessage()` (default = false) that attempts to center the text based on message length (bounding boxes are not available for textboards, so can't use those). The scaling factor is a bit eyeballed, but it seems to work for the text sizes I currently use.

### DIFF
--- a/src/main/kotlin/graphics/scenery/Camera.kt
+++ b/src/main/kotlin/graphics/scenery/Camera.kt
@@ -359,8 +359,8 @@ open class Camera : DefaultNode("Camera"), HasRenderable, HasMaterial, HasCustom
                 var maxX = 0f
                 while (bv.hasRemaining()) {
                     maxX = java.lang.Float.max(bv.get(), maxX)
-                    bv.get()
-                    bv.get()
+                    bv.get() // skip Y
+                    bv.get() // skip Z
                 }
                 maxX *= tb.spatial().scale.x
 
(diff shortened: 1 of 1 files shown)


src/main/kotlin/graphics/scenery/Camera.kt
```diff
@@ -336,14 +336,30 @@ open class Camera : DefaultNode("Camera"), HasRenderable, HasMaterial, HasCustom
...
+        backgroundColor: Vector4f = Vector4f(0.0f),
+        duration: Int = 3000,
+        centered: Boolean = false
+    ) {
         val tb = TextBoard()
         tb.fontColor = messageColor
         tb.backgroundColor = backgroundColor
         tb.text = message
+        // Calculate the offset for text centering based on message length.
+        // The second divisor was eyeballed and might need adjustment.
```
moreApi, line 354: if you do it like this https://github.com/moreApi/microscenery/blob/52b2e5156609f8489c9fb053d508db9ec12170d2/frontend/src/main/kotlin/microscenery/VRUI/Gui3D/TextBox.kt#L57 you dont need to guess :)

moreApi (changes requested):

src/main/kotlin/graphics/scenery/Camera.kt
smlpt: Good suggestion, I changed it 👍

moreApi: sorry I should have been clearer. the part before the linked line is also relevant. I basically check for the right most point of the geometry and then apply the scale.

smlpt: this has been fixed now

moreApi (approved):
lgtm

skalarproduktraum (approved):
lgtmt! Thanks, @smlpt 👍

PR #776, opened 2024-09-12
Sample ray grid traversal

smlpt:
This PR adds another volume sampling function in addition to the existing sample ray function. The downside of sampling a volume at even intervals is that some voxels may be sampled twice, resulting in plateaus that can make it difficult to find local maxima.
This is why I implemented the grid traversal algorithm by [Amanatides and Woo](http://www.cse.yorku.ca/~amana/research/grid.pdf), which also takes an entry and an exit position, but it traverses the grid one voxel at a time instead of moving directly along the ray at even step sizes.

Because there are no evenly sized delta values to return here, this function returns two lists as a pair; one list that contains all sampled values, and one list with all corresponding voxel positions.


smlpt:
This PR now also fixes several issues with RAI volume sampling and buffered volume sampling, especially with anisotropic volumes.

kephale (approved):
I read this over. It looks quite reasonable.

smlpt:
Thanks @kephale! It looks like I can't rebase it due to conflicts, but Github won't show me which ones. Shall we merge instead?

smlpt:
@kephale I think this PR will also fix https://github.com/scenerygraphics/sciview/issues/605, once we publish a new scenery release.

kephale:
@skalarproduktraum bump

skalarproduktraum (approved):
Thanks @smlpt 👍

PR #780, opened 2024-09-24
DSSDO: Fix wrong attachment order for samples

skalarproduktraum:
This PR fixes an issue in the DSSDO shaders introduced in #594. There, the order of sampled attachments was wrong, causing the DSSDO shader to sample depth from the emissions texture, leading to visual artifacts.


moreApi (approved):
lgtm, but I dont know much about the shaders

PR #785, opened 2024-10-30
Improve naming of hasTransferFunction.range property to make its function clear and avoid confusion with minDisplayRange and maxDisplayRange

moreApi:
Improve naming of hasTransferFunction.range property to make its function clear and avoid confusion with minDisplayRange and maxDisplayRange


smlpt (approved):
LGTM!

skalarproduktraum (approved):
Thanks @moreApi 👍

PR #786, opened 2024-11-14
Bump scijava parent POM to 39.0.0

skalarproduktraum:
This PR bumps the dependency on the scijava parent POM to 39.0.0, and also bumps Kotlin to 2.0.21, while removing a spurious import from RenderConfigReader.


kephale (approved):
LGTM!

PR #788, opened 2024-11-18
OpenVRHMD safe shutdown

smlpt:
This PR adds `initalized = false` to `OpenVRHMD.close()` before calling the shutdown logic to prevent the renderer from accessing memory that doesn't exist anymore.


skalarproduktraum:
@smlpt While you are at it, would you be so kind to check the other functions in OpenVRHMD as well?

smlpt:
> @smlpt While you are at it, would you be so kind to check the other functions in OpenVRHMD as well?

Except for various non-nullable methods that access HMD properties, nothing comes to mind.
You mentioned previously you'd like to eventually re-think the whole shutdown logic, but I think that's a story for another PR.

PR #792, opened 2025-01-16
3D GUI updates

smlpt:
This PR adds VR GUI functionality from [microscenery](https://github.com/moreApi/microscenery) and [jans branch](https://github.com/scenerygraphics/scenery/tree/jans-branch), and tweaks it to work without further dependencies.
This includes UI elements like Button, Tabbedmenu, Switch and Textbox, as well as layout elements like Column and Row.

It also updates VRTouch, VRGrab and VRPRess interactions from @moreApi's branch.
Further updates: VRTouch objects can now be interacted with custom nodes (think tool tips) instead of just the controller hitbox. Buttons can now optionally be triggered by touch only, not necessarily requiring VRPress behavior anymore, and custom colors and pressedColors can be passed to them.

### DIFF
--- a/src/main/kotlin/graphics/scenery/ui/Button.kt
+++ b/src/main/kotlin/graphics/scenery/ui/Button.kt
@@ -83,7 +83,9 @@ open class Button(
             },
             onHold = {
                 // this prevents immediate on/off switching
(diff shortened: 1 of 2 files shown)


moreApi:
I unsibscribed from this PR because I got an notification for each little commit. Just add or poke me again, when my attention is required again.

skalarproduktraum (approved):

PR #798, opened 2025-08-26
Fix volume render speeds

smlpt:
This PR speeds up volume rendering by extracting the textureSize call from the sampling loop and putting it into a uniform that is filled in the VolumeManager class. 

To be able to test it, this branch was branched from https://github.com/scenerygraphics/scenery/pull/795.

With a simple T1 head example, this speeds up RAI volume rendering from 85 to 160 fps.
In more complex scenarios (1.5GB volume), the speed-up is 10 fps, from 60 to 70fps.


smlpt:
I wonder if we should keep this PR open to investigate further rendering speedup potentials, since Nsight Graphics still reports major long scoreboard stalls from fetching the volume texture.

skalarproduktraum:
@smlpt on second thought – just remove the changes in `Volume` in this PR, then it's good to go 🚀

PR #799, opened 2025-09-02
New Sonatype publishing & Kotlin update

skalarproduktraum:
This PR adds support for the new way of publishing artifacts on Sonatype. This works now by running `./gradlew -Prelease=true publishAggregationToCentralPortal` (set `release` to false for a SNAPSHOT). The artifacts are not automatically published but need to be released via the Sonatype portal.

The PR bumps Kotlin to 2.2.10, the latest available version.
The PR also bumps the version of project to 0.12.0 (which I already released by accident 🤦‍♂️).


smlpt (approved):
looks good!

PR #815, opened 2025-11-20
Fix mouse button ordering across platforms

ctrueden:
This PR fixes the inconsistency in the handling of mouse buttons across platforms.

The first commit enhances the MouseInputExample to map all three mouse buttons to different verifiable behaviors in the scene, and renumbers the keyboard "sythetic mouse button" equivalents to align with the mouse's physical ordering: 1=left, 2=middle, 3=right—and then for double-click+drag actions, 4=left, 5=right. This change makes it very easy to test if the mouse button scenery thinks it's mapping in each case actually corresponds to reality.

The second commit actually fixes the mouse button ordering bug; see commit message for details.

I tested these changes on Linux, and can verify that before the bug-fix commit, the middle and right mouse buttons are unexpectedly swapped, behavior-wise, whereas after the fix, they both work as advertised.

Edit: I have now tested on macOS as well, and confirm that both *before* and *after* the bug-fix commit, the middle and right mouse buttons are *correct*. So this PR has no negative impact on the macOS platform. All that's left is to test on Windows, then.

As an aside: on macOS I see an all-gray canvas rather than the scene a large majority of the time—and this happens with both this branch (up-to-date on top of current `main`) as well as these commits cherry-picked onto `scenery-0.10.0`—but for a small fraction of launches, the scene appears and can be interacted with, which is how I was able to verify this bug-fix there.

### DIFF
--- a/src/main/kotlin/graphics/scenery/controls/GLFWMouseAndKeyHandler.kt
+++ b/src/main/kotlin/graphics/scenery/controls/GLFWMouseAndKeyHandler.kt
@@ -64,10 +64,10 @@ open class GLFWMouseAndKeyHandler(var hub: Hub?) : MouseAndKeyHandlerBase(), Aut
             mask = mask or InputEvent.BUTTON1_DOWN_MASK
         }
         if(buttonKey == GLFW_MOUSE_BUTTON_RIGHT) {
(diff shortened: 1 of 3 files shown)


moreApi (approved):
it also works on windows.

thank you!

PR #822, opened 2026-02-19
Bump lwjgl to 3.4.1, pin lwjgl-openvr to 3.3.6

smlpt:
Newer Nvidia drivers conflict with vulkan initialization in LWJGL 3.3.3 (see http://forum.lwjgl.org/index.php?topic=7334.0). This has been [fixed in 3.4.0](https://github.com/LWJGL/lwjgl3/commit/b5b363e2).

The LWJGL-OpenVR bindings however no longer ship with 3.4.0+, so these are pinned to 3.3.6 for now.

Long-term a rewrite to OpenXR might be required.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,7 +28,7 @@ val lwjglArtifacts = listOf(
         "lwjgl-jemalloc",
         "lwjgl-vulkan",
         "lwjgl-opengl",
-        "lwjgl-openvr",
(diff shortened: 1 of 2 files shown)


smlpt:
Closing for now since separating the OpenVR versions from the remaining artifacts is a bad idea.
Will need a full refactor to OpenXR.

PR #823, opened 2026-04-23
manvr3d improvement sprint v0.4

smlpt:
Various bug fixes and changes, as needed by https://github.com/scenerygraphics/manvr3d/pull/30
Current changes:
- 3D cursor now reflects manvr3d's sphere scale factor
- support for volume lensing in the shaders
- adds example file for volume lensing API
- Cursortool now has a pointer cylinder that can be toggled from there spherical cursor for better UI interaction
- Fix MultiWristMenu bounding box generation


src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
```diff
@@ -389,6 +389,11 @@ open class VulkanRenderer(hub: Hub,
     protected var heartbeatTimer = Timer()
     protected var gpuStats: GPUStats? = null
 
+    val textureDispatcherJob = SupervisorJob()
```
skalarproduktraum: I don't see why this should be here. Shouldn't this be part of VulkanTexture instead?

src/main/kotlin/graphics/scenery/Mesh.kt
skalarproduktraum, line 1: Those info logs were actually left there deliberately, so it's clear what's being imported 😅

skalarproduktraum (changes requested):
Responsibility issue in VulkanRenderer, please fix that. Otherwise looking good 👍

src/main/kotlin/graphics/scenery/Mesh.kt
smlpt: Okay, I can leave them as info logs. They cluttered the manvr3d console a lot, but I can switch the runtime log settings to only show warnings from scenery, not info logs.

src/main/kotlin/graphics/scenery/backends/vulkan/VulkanRenderer.kt
smlpt: moved to VulkanTexture. Looks like everything still works.

smlpt:
@skalarproduktraum changes added, could you pls have a look again?

skalarproduktraum (approved):
lgtm, thanks @smlpt 👍

PR #824, opened 2026-05-06
Bump to BVV 0.4.1

smlpt:
This PR adds support for BigVolumeViewer 0.4.1 in scenery.


smlpt:
@ekatrukha Thanks for the work on this! I tested your branch and it seems to work fine. Could you please remove the hardcoded path in the BDVExample? Then @skalarproduktraum can merge it.

ekatrukha:
Hello @smlpt adn @skalarproduktraum,

I didn't finish it completely. 

I think it works only with 16-bit inputs now, since it uses only one cache and PBO chain. 
In principle, one needs two separate ones (like in BVV now), for 8bit and 16bit, it should be relatively simple.
But I am a bit slow with Kotlin, didn't use it much. 
If time allows, I will work on it tomorrow (+ remove the hardcoded example, and try to bring it up to the current state).

smlpt:
> Hello @smlpt adn @skalarproduktraum,
> 
> I didn't finish it completely.
> 
> I think it works only with 16-bit inputs now, since it uses only one cache and PBO chain. In principle, one needs two separate ones (like in BVV now), for 8bit and 16bit, it should be relatively simple. But I am a bit slow with Kotlin, didn't use it much. If time allows, I will work on it tomorrow (+ remove the hardcoded example, and try to bring it up to the current state).

Thank you! There's no big rush on our side, so if its sometime later this month that would also work.

ekatrukha:
Hello @smlpt and @skalarproduktraum,

I've added separate caches for 8bit and 16bit multires images. Basically, the logic is very similar to changes in BVV's [VolumeRenderer](https://github.com/bigdataviewer/bigvolumeviewer-core/blob/master/src/main/java/bvv/core/render/VolumeRenderer.java).

I've tested this branch on a single 8-bit or 16-bit multires volume, and it seems to work fine. 
1) Can you please check that it works if one adds both 8bit and 16bit simultaneously? It should be easy to make 8bit BDV XML of `t1-head`.
2) And can you please check if adding one or the other volume later also works?

Cheers,
Eugene

PR #826, opened 2026-09-22
SwingSwapchain: harden canvas/window setup against intermittent blank rendering on macOS

moreApi:
What

Hardens SwingSwapchain's window/canvas setup on macOS to reduce (but not fully eliminate — see linked issue) an intermittent bug where the render window opens but shows no content, even though the renderer is running and presenting frames successfully underneath.

Changes in SwingSwapchain.createWindow():

canvas.ignoreRepaint = true, so AWT's own repaint machinery can never paint over the canvas's Metal-driven content with its plain background.
p.validate() synchronously after adding the canvas, instead of relying on a layout pass being deferred to the EDT.
frame.pack() again after the canvas is added — per LWJGL's own AWTVK docs, this is required to force a real native peer for the canvas before AWTVK.create() is called; it was previously only being called once, before the canvas existed.
The resize listener (swapchainRecreator.mustRecreate = true on resize) is now attached to canvas in addition to p, since canvas's bounds — set by p's BorderLayout — can change independently of p's own size.
A bounded (2s) readiness wait for the canvas to be showing, bound to a real GraphicsConfiguration, and have a non-zero size, before creating the Vulkan surface from it.
Why

Full writeup with the investigation and evidence is in #827 — short version: the actual root cause looks like a race in LWJGL's AWTVK/CAMetalLayer bridge on Apple Silicon macOS (confirmed via a VK_TIMEOUT from MoltenVK itself showing presented frames are sometimes never acknowledged by the compositor), which isn't fixable from scenery's side. These changes are legitimate, independently-justified fixes to SwingSwapchain's AWT setup that were missing regardless, and they do appear to reduce how often the underlying race is lost — but they're not a complete fix. Filing that separately upstream.

Note

Full credit for the investigation here goes to Claude (Anthropic's Claude Code) — I ran the reproductions, pasted back logs/console output, and reported what I observed on-screen, but the ruling-out of each hypothesis (AWT readiness, Java2D pipeline, async texture races, Vulkan acquire/present errors) and the actual diagnosis (the MTL_HUD_ENABLED test, the forced-recreate VK_TIMEOUT finding) were Claude's. Flagging that for transparency.

Test plan
 Built and ran TexturedCubeExample repeatedly (bare reruns, no clean) on Apple Silicon macOS; content renders more reliably than before, though not 100% of the time.
 No regressions on other platforms (this only touches the macOS-specific SwingSwapchain path).


moreApi:
Build is broken not because of changes of this PR

